PR #2, opened 2018-04-18
T2D-155 Generalise Metadata API

Zhicheng-Liu:


src/main/java/uk/ac/ebi/ampt2d/metadata/SwaggerApiInfoProperties.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public String getLicenseUrl() {
+        return licenseUrl;
+    }
+
+    public void setLicenseUrl(String licenseUrl) {
+        this.licenseUrl = licenseUrl;
+    }
+
+    public class Contact {
```
jorizci: If it is public, separate file.

src/main/java/uk/ac/ebi/ampt2d/metadata/properties/SwaggerApiInfoProperties.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @NotBlank(message="API title can not be blank.")
+    private String title;
+
+    private String description;
+
+    @NotBlank(message="API version can not be blank.")
+    private String version;
+
+    @URL(message="Terms of Service URL is not valid.")
```
cyenyxe: Does this accept a blank URL? Judging by the test properties file, it seems so, but just to be sure.

Zhicheng-Liu: Yes, it will accept a blank URL. And the swagger-ui will not show the link to the terms of service if the URL is blank or not defined.
If it is not blank, then the value will be validated against @URL constraint.

PR #6, opened 2018-05-08
T2D-156 Add accessions to metadata objects

Zhicheng-Liu:


src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Study.java
```diff
@@ -38,11 +34,12 @@
...
-    @ApiModelProperty(position = 1, value = "Study auto generated id", required = true, readOnly = true)
-    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
+    @ApiModelProperty(position = 1, required = true)
+    @Size(min = 1, max = 255)
+    @NotNull
+    @JsonProperty
     @Id
-    @GeneratedValue(strategy = GenerationType.AUTO)
-    private long id;
+    private String accession;
```
selvaebi: as we have versions in accessions , should we have accessions as id ?

Zhicheng-Liu: Normally a versioned accession could be represented as "$accession.$version". I think we can use that as the ID. What do you think? @cyenyxe and @jorizci

jorizci: As last days have proved this could be more complex, imho, Id and versions should not be linked. If we have to support versioning or history of the accession then we have to devise a different structure. Or procedure. It can be an independant id and a indexed accession column plus a version column.

I don't recall what is specified on the user stories. @cyenyxe?

selvaebi:
Please review the last commit only

selvaebi:
Tried composite keys , but we can access links with single id only like accession.version. if we need to search only by providing accessions we need to write a custom controller( same with both custom id and composite pks ) .

src/main/java/uk/ac/ebi/ampt2d/metadata/controllers/StudyRestController.java
```diff
@@ -58,20 +61,22 @@
...
+    @RequestMapping(method = RequestMethod.GET, path = "/search/text")
+    public Resources<Study> getStudies(@RequestParam("searchString") String searchString) {
         QStudy study = QStudy.study;
         Predicate predicate = study.name.containsIgnoreCase(searchString).
                 or(study.description.containsIgnoreCase(searchString));
-        return studyRepository.findAll(predicate);
+        Iterable<Study> studyIterable = studyRepository.findAll(predicate);
+        return new Resources(studyIterable, ((List<Study>) studyIterable).stream().
+                map(studyObj -> ControllerLinkBuilder.linkTo(Study.class).slash("studies").slash(studyObj.getId())
+                        .withRel("search")).collect(Collectors.toList()));
```
cyenyxe: Remove blank line.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/BaseAccessionVersionEntityId.java
```diff
@@ -0,0 +1,74 @@
...
+package uk.ac.ebi.ampt2d.metadata.persistence.entities;
+
+import javax.persistence.Embeddable;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+import java.io.Serializable;
+
+@Embeddable
+public class BaseAccessionVersionEntityId implements Serializable {
```
cyenyxe: `AccessionVersionEntityId` or `MetadataEntityId`, but the "base" part means nothing to me at least.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/File.java
```diff
@@ -43,24 +43,30 @@
     }
 
     @ApiModelProperty(position = 1, required = true)
+    @Valid
+    @EmbeddedId
+    public BaseAccessionVersionEntityId id;
+
```
cyenyxe: Remove blank line.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/File.java
```diff
@@ -43,24 +43,30 @@
...
+    @ApiModelProperty(position = 2, required = true)
     @NotNull
     @Size(min = 1, max = 255)
     @JsonProperty
-    @Id
+    @Column(nullable = false)
     private String hash;
 
-    @ApiModelProperty(position = 2, required = true)
+    @ApiModelProperty(position = 3, required = true)
```
cyenyxe: Do we really need the position to be declared? Every time we change the attributes these numbers need to be modified too.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/File.java
```diff
@@ -43,24 +43,30 @@
...
     @Size(min = 1, max = 255)
     @JsonProperty
     @Column(nullable = false)
-    private String fileName;
+    private String name;
 
-    @ApiModelProperty(position = 3, required = true)
+    @ApiModelProperty(position = 4, required = true)
     @JsonProperty
     private long fileSize;
```
cyenyxe: `size` instead of `fileSize` because it is redundant with the class name.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/idconverter/CustomBackendIdConverter.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.springframework.core.convert.converter.Converter;
+import uk.ac.ebi.ampt2d.metadata.persistence.entities.BaseAccessionVersionEntityId;
+
+public class CustomBackendIdConverter implements Converter<String, BaseAccessionVersionEntityId> {
+
+    @Override
+    public BaseAccessionVersionEntityId convert(String id) {
+        if (id.split("\\.").length != 2)
+            throw new IllegalArgumentException("Please provide id in the pattern accession.version");
```
cyenyxe: Is this sentence correct? What about the following?

> Please provide an ID of the form "accession.version"

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/idconverter/CustomBackendIdConverter.java
```diff
@@ -0,0 +1,31 @@
...
+        return new BaseAccessionVersionEntityId(id.split("\\.")[0], Integer.parseInt(id.split("\\.")[1]));
```
cyenyxe: This method splits the same string 3 times. Do it just once and save it in a variable.

src/test/java/uk/ac/ebi/ampt2d/metadata/MetadataApplicationTest.java
```diff
@@ -173,11 +175,12 @@ private String postTestStudy(String name) throws Exception {
     public void postAnalysis() throws Exception {
         String assemblyUrl = postTestAssembly("GRCh37", "p2",
                 Arrays.asList("GCA_000001405.3", "GCF_000001405.14"));
-        String studyUrl = postTestStudy("test human study");
+        String studyUrl = postTestStudy("EGAS0001", 1, "test_human_study");
 
         MvcResult mvcResult = mockMvc.perform(post("/analyses")
                 .content("{ " +
-                        "\"name\": \"test human analysis\"," +
+                        "\"id\":{ \"accession\": \"EGAA0001\",\"version\": " + 1 + "}," +
```
cyenyxe: If "1" is not coming from a constant, write it inside the string to avoid unnecessary concatenations.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/File.java
selvaebi: yes . Else the parameters is displayed by alphabetical order.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/idconverter/CustomBackendIdConverter.java
jorizci: Aditionally, parseInt can throw a parsing exception if the second field is not a number. You should control that case and give the same response as before.

src/main/java/uk/ac/ebi/ampt2d/metadata/controllers/StudyRestController.java
```diff
@@ -57,21 +61,35 @@
...
 
+    @ApiOperation(value = "Get the latest version of study based on accession")
+    @RequestMapping(method = RequestMethod.GET, path = "/search/accession", produces = "application/json")
+    @ResponseBody
+    public Resources<Study> searchByAccession(@RequestParam("accession") String accession) {
+        QStudy study = QStudy.study;
+        Predicate predicate = study.id.accession.eq(accession).and(study.id.version.eq(JPAExpressions.
+                select(study.id.version.max()).from(study).where(study.id.accession.eq(accession))));
+        Iterable<Study> studyIterable = studyRepository.findAll(predicate);
+        return new Resources(studyIterable, ((List<Study>) studyIterable).stream().
```
selvaebi: Everything goes into Studyservice once filter studies by taxonomy is merged

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/File.java
cyenyxe: And what is the problem with that? It's not like users are going to look into the code and be angry at the mismatch.

PR #14, opened 2018-07-04
T2D-193 Deprecate study

Zhicheng-Liu:
Add a boolean field to indicate if a study has been deprecated. Deprecated
studies will not be returned in any public API.

### DIFF
--- a/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/StudyDeprecationAspect.java
+++ b/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/StudyDeprecationAspect.java
@@ -41,7 +41,7 @@ public class StudyDeprecationAspect {
     }
 
     @Around("execution(* uk.ac.ebi.ampt2d.metadata.persistence.repositories.StudyRepository.findOne(..))")
-    public Object filterOnDeprecadtionAdviceFindOne(ProceedingJoinPoint proceedingJoinPoint) throws Throwable {
+    public Object filterOnDeprecationAdviceFindOne(ProceedingJoinPoint proceedingJoinPoint) throws Throwable {
         Object study = proceedingJoinPoint.proceed();
 
         if ( study != null && ((Study) study).isDeprecated() ) {
--- a/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/StudyDeprecationAspect.java
+++ b/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/StudyDeprecationAspect.java
@@ -44,7 +44,7 @@ public class StudyDeprecationAspect {
     public Object filterOnDeprecationAdviceFindOne(ProceedingJoinPoint proceedingJoinPoint) throws Throwable {
         Object study = proceedingJoinPoint.proceed();
 
-        if ( study != null && ((Study) study).isDeprecated() ) {
+        if ( study == null || ((Study) study).isDeprecated() ) {
             return null;
         }
 
--- a/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/StudyDeprecationAspect.java
+++ b/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/StudyDeprecationAspect.java
@@ -1,0 +1,54 @@
+/*
+ *
+ * Copyright 2018 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package uk.ac.ebi.ampt2d.metadata.aop;
+
+import com.querydsl.core.types.Predicate;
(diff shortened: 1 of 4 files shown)


selvaebi:
Dont we need a endpoint for deprecating a study ?

Zhicheng-Liu:
@selvaebi you can update a study with different `deprecated` value via `PATCH` request. The endpoint is provided by default by Spring Data Rest.

selvaebi:
ya the test cases was using post only , can we change it to patch to make it meaningful. and also make the deprecated flag default as false.

Zhicheng-Liu:
I see. I will add test cases in which an existing study could be deprecated.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Study.java
```diff
@@ -66,6 +66,7 @@
     @ManyToOne(optional = false)
     private Taxonomy taxonomy;
 
+    @NotNull
     @JsonProperty
```
selvaebi: I think we can add  @JsonProperty(defaultValue = "false")

Zhicheng-Liu: OK! I have reverted my last commit for adding `@NotNull` here. The annotation does not work with primitive types.
Because this field `deprecated` is a `boolean`, a primitive type so it has default value as `false` anyway.
The `@JsonProperty(defaultValue = "false")` does not set the default value for the field. It is there mostly for "documentation" purpose only, as explained in https://fasterxml.github.io/jackson-annotations/javadoc/2.9/com/fasterxml/jackson/annotation/JsonProperty.html#defaultValue()

So I am happy with `boolean deprecated`. But I can add `defaultValue = "false"` to `@JsonProperty` for better documentation.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Study.java
```diff
@@ -66,6 +66,10 @@
     @ManyToOne(optional = false)
     private Taxonomy taxonomy;
 
+    @JsonProperty(defaultValue = "false")
```
cyenyxe: If we are not going to display studies that have been deprecated, I found find more natural to annotate this with `@JsonIgnore`.

selvaebi: I think if we put jsonignore we cant patch , if we wanna deprecate, and also we need to include a seperate endpoint to deprecate / undeprecate .

src/main/java/uk/ac/ebi/ampt2d/metadata/aop/StudyDeprecationAspect.java
```diff
@@ -0,0 +1,54 @@
...
+        args[0] = deprecationCheckedPredicate;
+
+        return proceedingJoinPoint.proceed(args);
+    }
+
+    @Around("execution(* uk.ac.ebi.ampt2d.metadata.persistence.repositories.StudyRepository.findOne(..))")
+    public Object filterOnDeprecationAdviceFindOne(ProceedingJoinPoint proceedingJoinPoint) throws Throwable {
+        Object study = proceedingJoinPoint.proceed();
+
+        if ( study != null && ((Study) study).isDeprecated() ) {
```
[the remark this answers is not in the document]
jorizci: Question, if we return null here what is the final response at the endpoint?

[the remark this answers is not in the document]
Zhicheng-Liu: @jorizci if we return null here, the final response will be "Not Found".

src/main/java/uk/ac/ebi/ampt2d/metadata/aop/StudyDeprecationAspect.java
```diff
@@ -0,0 +1,54 @@
...
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import uk.ac.ebi.ampt2d.metadata.persistence.entities.QStudy;
+import uk.ac.ebi.ampt2d.metadata.persistence.entities.Study;
+
+@Aspect
+public class StudyDeprecationAspect {
+
+    @Around("execution(* uk.ac.ebi.ampt2d.metadata.persistence.repositories.StudyRepository.findAll(..))")
```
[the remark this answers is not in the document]
Zhicheng-Liu: Indeed, `Before` sounds more logical than `Around` here, so is `AfterReturning` to the `findOne` advice.  But due to the interface and implementation details of Spring AOP:
1). `Before` advice does not allow you to change the arguments of the join point method. It allows you to get the arguments, good for logging or other readonly purposes, but not for the case here in which we have to add an additional criteria to its arguments.
2). `Around` advice allows you to get the arguments of the join point method, do whatever you want and then call `proceed` with the new arguments.
3). In `AfterReturning` advice  you can get the returned object from the execution of join point method, but you are not allowed to change it. While in `Around` advice you can inspect the returned object and change it.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Study.java
Zhicheng-Liu: OK! I have added in commit 8b0b4e7 a `access` property for `@JsonProperty` annotation. It is available from recent Jackson library. I have set it to `WRITE_ONLY` so that we can do `POST` and `PATCH` as @selvaebi mentioned, but also in the repsonse jsons of `GET` requests, the `deprecated` field will not appear. I think that is what @cyenyxe wanted, right?

PR #15, opened 2018-07-06
T2D-191 Add lastModifiedDate to metadata objects

Zhicheng-Liu:


src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Auditable.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.data.annotation.LastModifiedDate;
+import org.springframework.data.jpa.domain.support.AuditingEntityListener;
+
+import javax.persistence.EntityListeners;
+import javax.persistence.MappedSuperclass;
+import java.time.ZonedDateTime;
+
+@MappedSuperclass
+@EntityListeners(AuditingEntityListener.class)
+abstract class Auditable {
```
cyenyxe: Maybe instead of creating this new class, `BaseEntity` could be changed from interface to class? That way it wouldn't be necessary to make changes to every single entity class.

Zhicheng-Liu: OK! I have merged `BaseEntity` and `Auditable`.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Auditable.java
```diff
@@ -17,8 +17,27 @@
...
+import javax.persistence.EntityListeners;
+import javax.persistence.MappedSuperclass;
+import java.time.ZonedDateTime;
 
+@MappedSuperclass
+@EntityListeners(AuditingEntityListener.class)
+public abstract class Auditable<ID> {
+
+    @LastModifiedDate
+    private ZonedDateTime lastModifiedDate;
```
selvaebi: Please check is there a way to display only on get request , not on post/patch request

Zhicheng-Liu: Fixed in 753d68e.

PR #18, opened 2018-07-24
T2D-197 Add endpoints for CRUD linked studies

Zhicheng-Liu:
Add a new table for the many-to-many relationships for linked studies.
And also add new service and controller for the endpoints.

PR #23, opened 2018-09-05
T2D-197 Find linked studies of a study

Zhicheng-Liu:


selvaebi:
I am not sure , if we link an un existing study it does not give us any error. Unlike if we post a un existing taxonomy it throws us error like taxonomy does not exist.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Study.java
```diff
@@ -83,6 +83,10 @@
     @Column
     private boolean browsable;
 
+    @JsonProperty
```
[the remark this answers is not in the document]
Zhicheng-Liu: Added in 191d51c.

Zhicheng-Liu:
@selvaebi : for the taxonomy it is a `ManyToOne` relationship so `taxonomy` is a column in table `study` and it has `NotNull` and foreign key constraints applied. For the `linkedStudies` here  there is no `NotNull` constraint so a nonexisted linked study will be treated as `null`.

cyenyxe:
That behavior is a bit unintuitive, is there any way to work around that?

Could you please also solve the conflicts in the test file?

Zhicheng-Liu:
@cyenyxe : I tried but there seemed not to be an obvious workaround. The closed pull request (https://github.com/EBIvariation/metadata-ws/pull/18) is a proper solution imho. 

I have resolved the conflicts in the test file, BTW.

selvaebi:
Please correct me if am wrong , if a study can be linked to more than one study , should the relationship be ManyToMany ?

Zhicheng-Liu:
The current implementation is more towards parent->children like relationship. We can't now link two studies to the same study, i.e. a "child" could not have two "parent", but a "parent" could have multiple "children". But this covers our use cases if I am not wrong.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Study.java
```diff
@@ -83,6 +83,11 @@
     @Column
     private boolean browsable;
 
+    @ApiModelProperty(position = 9)
+    @JsonProperty
+    @OneToMany
+    private List<Study> linkedStudies;
```
cyenyxe: Given these should be the children studies only, we should the attribute that way. Otherwise it will sound like the `StudyService.findLinkedStudies` method simply returns the value of this attribute, when it is not the case.

Zhicheng-Liu: OK! I have renamed the attribute to `childStudies` as discussed in 07d0b24

PR #24, opened 2018-09-06
T2D-225 keycloak auth

selvaebi:


selvaebi:
Application is accessed via access token retrieved from below request ,

curl --data "grant_type=password&client_id=ampt2d-metadata&client_secret=cea914bc-91e4-4774-ad9d-0ea84ba2cce9&username=selva.439@gmail.com&password=selva" https://www.ebi.ac.uk/ega/ampt2d/auth/realms/Ampt2d/protocol/openid-connect/token 

the access token is used as bearer token in request header. In swagger ui the token should be passed as "Bearer <token>" . 

additional users can be added in http://ampt2d-gait.windows.ebi.ac.uk:8880/auth/ ( the proxy url only works for getting token , the admin console need to be configured properly behind proxy to be accessible via proxy)

PR #26, opened 2018-09-12
T2D-217 paging and sorting repository query

selvaebi:


selvaebi:
Hi Cristina , 
Yes it still works as we are using StudyRepository.findAll(..) ( which will match  findAll(Predicate,Pageable) of QueryDslPredicateExecutor )  
and direct CrudRepository.findOne(Study).

PR #27, opened 2018-09-19
T2D-220 Make Assembly path not nullable

Zhicheng-Liu:


Zhicheng-Liu:
It turns out the `patch` could be empty as described in @cyenyxe 's comment above. Close this PR as it is no longer needed.

PR #28, opened 2018-09-26
T2D-223 Web resource accepts an empty URL

iyangar:
Fixes

### DIFF
--- a/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/WebResource.java
+++ b/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/WebResource.java
@@ -20,7 +20,6 @@ package uk.ac.ebi.ampt2d.metadata.persistence.entities;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.annotations.ApiModelProperty;
 import org.hibernate.validator.constraints.NotEmpty;
-import org.hibernate.validator.constraints.URL;
 
 import javax.persistence.Column;
 import javax.persistence.Entity;
@@ -31,7 +30,6 @@ import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Pattern;
-import java.lang.annotation.ElementType;
 
 @Entity
 public class WebResource extends Auditable<Long> {
@@ -64,7 +62,7 @@ public class WebResource extends Auditable<Long> {
     @JsonProperty
     @Column(nullable = false, columnDefinition = "TEXT")
     @NotEmpty
-    @Pattern(message = "Must be a valid URL.", regexp="(^(https?|ftp):)?(//|\\\\)[-a-zA-Z0-9+&@#/%?=~_|!:,.;$'*\\[\\]()]*")
+    @Pattern(message = "Must be a valid URL.", regexp="(^(https?|ftp):(//|\\\\))[-a-zA-Z0-9+&@#/%?=~_|!:,.;$'*\\[\\]()]+")
     private String resourceUrl;
 
     WebResource() {}
--- a/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/WebResource.java
+++ b/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/WebResource.java
@@ -19,6 +19,7 @@ package uk.ac.ebi.ampt2d.metadata.persistence.entities;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.annotations.ApiModelProperty;
+import org.hibernate.validator.constraints.NotEmpty;
 import org.hibernate.validator.constraints.URL;
 
 import javax.persistence.Column;
@@ -29,6 +30,7 @@ import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
 
 @Entity
 public class WebResource extends Auditable<Long> {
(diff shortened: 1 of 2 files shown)


src/test/java/uk/ac/ebi/ampt2d/metadata/MetadataApplicationTest.java
```diff
@@ -305,6 +305,15 @@ private String postTestWebResource() throws Exception {
         return mvcResult.getResponse().getHeader("Location");
     }
 
+    @Test
+    public void postWebResourceNullURL() throws Exception {
```
cyenyxe: Could you please add another test that fails for a non-empty, but still invalid URL?

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/WebResource.java
```diff
@@ -61,6 +63,8 @@
     @URL
     @JsonProperty
     @Column(nullable = false, columnDefinition = "TEXT")
+    @NotEmpty
+    @Pattern(message = "Must be a valid URL.", regexp="^(https?|ftp|file):(//|\\\\)[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]")
```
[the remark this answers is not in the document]
iyangar: Sure, i have removed "file" protocol and its gets blocked also. But existing @URL code bars no-protocol URLs. If we remove @URL we will loose JSR validations for URL.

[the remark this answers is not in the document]
Zhicheng-Liu: Ah, I forgot there is a `@URL` annotation. It seems that `@URL` validation can take a few optional settings, like `protocol` and `regexp`: https://docs.jboss.org/hibernate/validator/5.4/api/org/hibernate/validator/constraints/URL.html

If these work for our purpose, I would suggest to use these instead of the `@Pattern` annotation.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/WebResource.java
```diff
@@ -58,9 +61,10 @@
 
     @ApiModelProperty(position = 3, required = true)
     @NotNull
-    @URL
```
Zhicheng-Liu: Could you please also removed the `import` for this?

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/WebResource.java
```diff
@@ -29,6 +30,8 @@
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
+import java.lang.annotation.ElementType;
```
Zhicheng-Liu: The `ElementType` is not used anywhere. It should not be left here.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/WebResource.java
```diff
@@ -58,9 +61,10 @@
 
     @ApiModelProperty(position = 3, required = true)
     @NotNull
-    @URL
     @JsonProperty
     @Column(nullable = false, columnDefinition = "TEXT")
+    @NotEmpty
+    @Pattern(message = "Must be a valid URL.", regexp="(^(https?|ftp):)?(//|\\\\)[-a-zA-Z0-9+&@#/%?=~_|!:,.;$'*\\[\\]()]*")
```
Zhicheng-Liu: The `regexp` is wrong. It will match `//www.ebi.ac.uk`.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/WebResource.java
iyangar: done

src/test/java/uk/ac/ebi/ampt2d/metadata/MetadataApplicationTest.java
```diff
@@ -305,6 +305,22 @@ private String postTestWebResource() throws Exception {
...
 
+    @Test
+    public void postWebResourceInvalidURL() throws Exception {
+        WebResource testWebResource = new WebResource(WebResource.Type.CENTER_WEB, "");
+
+        mockMvc.perform(post("/webResources")
+                .content(testWebResourceJson.write(testWebResource).getJson()))
+                .andExpect(status().is4xxClientError());
+
+        testWebResource = new WebResource(WebResource.Type.CENTER_WEB, "https://www.ebi.ac.uk<>");
```
cyenyxe: Could you please add some more tests to check valid and invalid URLs? For instance, making sure that only HTTP, HTTPS and FTP protocols are considered valid, as mentioned during @Zhicheng-Liu's review. Testing arbitrary errors in the URL would be also a good addition.

src/test/java/uk/ac/ebi/ampt2d/metadata/MetadataApplicationTest.java
```diff
@@ -305,6 +333,81 @@ private String postTestWebResource() throws Exception {
         return mvcResult.getResponse().getHeader("Location");
     }
 
+    private String postTestHTTPWebResource() throws Exception {
```
Zhicheng-Liu: You can group these four "repeated" `private` functions `postTest*WebResource` into one function taking one argument for the URLs with different protocols. Even further, you can also perform the `get` and `andExpect` checks in the same function.

src/test/java/uk/ac/ebi/ampt2d/metadata/MetadataApplicationTest.java
```diff
@@ -305,6 +333,81 @@ private String postTestWebResource() throws Exception {
...
+                .content(testWebResourceJson.write(testWebResource).getJson()))
+                .andExpect(status().is4xxClientError());
+
+        testWebResource = new WebResource(WebResource.Type.CENTER_WEB, "//fileserver/code/src/main/app.java");
+
+        mockMvc.perform(post("/webResources")
+                .content(testWebResourceJson.write(testWebResource).getJson()))
+                .andExpect(status().is4xxClientError());
+
+
```
Zhicheng-Liu: Personally I would not leave that many blank lines here. There is a coding style guide here for your reference: https://www.ebi.ac.uk/seqdb/confluence/display/VAR/Java+coding+style

PR #29, opened 2018-10-01
T2D-240 Add CORS config for Metadata API

Zhicheng-Liu:


src/test/java/uk/ac/ebi/ampt2d/metadata/MetadataApplicationTest.java
```diff
@@ -1295,4 +1298,14 @@ public void findLinkedStudies() throws Exception {
                 .andExpect(jsonPath("$..studies.length()").value(0));
     }
 
+    @Test
+    public void testCORS() throws Exception {
+        mockMvc.perform(options("/")
+                .header("Access-Control-Request-Method", "GET")
+                .header("Origin", "http://www.evil-url.com"))
```
selvaebi: evil url 😂

selvaebi:
If any other browser application gonna use the api outside ebi ? then origin * is fine . else we can allow only ebi.ac.uk. Other than that config looks fine to me .

cyenyxe:
There is no particular reason why should only be accessible from EBI servers. Would it be possible to deploy it in a wwwdev for testing from websites such as https://www.test-cors.org? Once we confirm with that one that is it okay, then I will approve this PR.

Zhicheng-Liu:
@cyenyxe I have deployed this pull request to our servers, can you have a look and check?

PR #30, opened 2018-10-02
T2D-222 made swagger to show up default value as 1 for version.

selvaebi:


selvaebi:
deployed to staging http://ves-hx-e9.ebi.ac.uk:8080/metadata/swagger-ui.html#/

PR #31, opened 2018-10-02
T2D-251 Updateddl and db

selvaebi:
RIght now as the ddl is validate , every time there is a change in schema we need to change the ddl to create and run the application locally to create the tables fresh, and then change the ddl to validate when deploying to envs. 
So to avoid the above manual create in local , we have to include a option while deploying to envs (create and validate) .

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,8 +16,8 @@ pipeline {
     productionHost = credentials('PRODUCTIONHOST')
   }
   parameters {
-    choice(choices: ['validate', 'update','create'], description: 'Behaviour at connection time for staging  only \
-    (initialize/update/validate schema)', name: 'dbBehaviour')
+    choice(choices: ['validate', 'update','create'], description: 'Behaviour at connection time for staging only \
+           (initialize/update/validate schema)', name: 'dbBehaviour')
     booleanParam(name: 'DeployToStaging' , defaultValue: false , description: '')
     booleanParam(name: 'DeployToProduction' , defaultValue: false , description: '')
   }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,7 +16,7 @@ pipeline {
     productionHost = credentials('PRODUCTIONHOST')
   }
   parameters {
-    choice(choices: ['validate', 'update','create'], description: 'Behaviour at connection time for staging  only
+    choice(choices: ['validate', 'update','create'], description: 'Behaviour at connection time for staging  only \
     (initialize/update/validate schema)', name: 'dbBehaviour')
     booleanParam(name: 'DeployToStaging' , defaultValue: false , description: '')
     booleanParam(name: 'DeployToProduction' , defaultValue: false , description: '')
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,13 +16,17 @@ pipeline {
     productionHost = credentials('PRODUCTIONHOST')
   }
   parameters {
+    choice(choices: ['validate', 'create'], description: 'Behaviour at connection time (initialize/validate schema)',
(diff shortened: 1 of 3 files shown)


selvaebi:
https://www.ebi.ac.uk/ega/ampt2d/jenkins/job/ampt2d-metadata/view/change-requests/job/PR-31/

selvaebi:
RIght now as the ddl is validate , every time there is a change in schema we need to change the ddl to create and run the application locally to create the tables fresh, and then change the ddl to validate when deploying to envs. 
So to avoid the above manual create in local , we have to include a option while deploying to envs (create and validate) .

selvaebi:
Thanks , the version update didn't affect the current unit test cases and jenkins file update can be tested while deployment using the above provided jenkins url.

Jenkinsfile
```diff
@@ -16,13 +16,17 @@ pipeline {
     productionHost = credentials('PRODUCTIONHOST')
   }
   parameters {
+    choice(choices: ['validate', 'create'], description: 'Behaviour at connection time (initialize/validate schema)',
```
Zhicheng-Liu: Can you add other choices according to https://docs.spring.io/spring-boot/docs/current/reference/html/howto-database-initialization.html. It would be good to maintain the order according to each choice's likelihood of being used.

selvaebi: I guess validate should be the only option , but to ease deployment for demo/testing purposes i added create.

Zhicheng-Liu: Although `validate` may be most frequently used, there are other choices like `update` (e.g. for adding a new column/constraint) could also be useful. What I meant was to make all avaialbe options for the `spring.jpa.hibernate.ddl-auto` a choice here. We may need one of them at some point in the future.

cyenyxe: I also wonder if all the options should be available in all the environments. For instance, using an option that would drop and recreate the production database could be **extremely** dangerous. It would be preferable to synchronize a staging database into production, and using only `validate` in production.

Zhicheng-Liu: If no better option out there, we can apply this choice only to staging DB and make production environments always use `validate`. I think `update` option is probably as useful as, if not more useful than, `create`for the staging. Sometime, there is only small "update" to the schema and you want to keep the data already loaded.

Zhicheng-Liu: Yes, that looks good to me.

selvaebi: i can do that but until we release or set up production ready metadata api ,
1. We will be testing the production only for demo purposes so we need to use the create for production to ease the deployment . 
2.Other option is make the staging public for testing purposes and implement the above ddl options in that way the prod ddl should be manually configured.
as far i understand option 2 looks good . i can make that way.

cyenyxe: Yes, the one we should use for demo/beta purposes is staging (wwwdev).

selvaebi:
https://www.ebi.ac.uk/ega/ampt2d/dev/metadata - dev/staging metadata url

Jenkinsfile
```diff
@@ -16,8 +16,8 @@ pipeline {
     productionHost = credentials('PRODUCTIONHOST')
   }
   parameters {
-    choice(choices: ['validate', 'create'], description: 'Behaviour at connection time (initialize/validate schema)',
-     name: 'dbBehaviour')
+    choice(choices: ['validate', 'update','create'], description: 'Behaviour at connection time for staging only \
+           (initialize/update/validate schema)', name: 'dbBehaviour')
```
cyenyxe: `ddlBehaviour` instead of `dbBehaviour` to make it more specific and match the name of the actual argument?

PR #36, opened 2018-10-22
T2D-218 added auto generated id and unique constraints

selvaebi:


src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/idconverter/CustomBackendIdConverter.java
```diff
@@ -18,19 +18,19 @@
 package uk.ac.ebi.ampt2d.metadata.persistence.idconverter;
 
 import org.springframework.core.convert.converter.Converter;
-import uk.ac.ebi.ampt2d.metadata.persistence.entities.AccessionVersionEntityId;
+import uk.ac.ebi.ampt2d.metadata.persistence.entities.AccessionVersionId;
 
-public class CustomBackendIdConverter implements Converter<String, AccessionVersionEntityId> {
+public class CustomBackendIdConverter implements Converter<String, AccessionVersionId> {
```
Zhicheng-Liu: With all the entities now use `long` ids, do we still need this id converter?

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/StudyServiceImpl.java
```diff
@@ -47,11 +46,9 @@ public Study findOneStudyByPredicate(Predicate predicate) {
     }
 
     @Override
-    public Study findOneStudyByAccession(AccessionVersionEntityId id) {
+    public Study findOneStudyById(Long id) {
```
Zhicheng-Liu: You can use `CrudRepository::findOne(ID id)` directly here now.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Taxonomy.java
```diff
@@ -31,14 +34,20 @@
...
+    private long id;
+
+    @ApiModelProperty(position = 2, example = "1")
     @NotNull
     @JsonProperty
-    @Id
     @Min(1)
-    private long id;
+    @Column(unique = true)
+    private long taxonomyIdentifier;
```
Zhicheng-Liu: I would prefer `taxonId` as it is shorter but still clear. It is just my opinion you don't have to change it.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/idconverter/CustomBackendIdConverter.java
selvaebi: it is used for querydsl queries , like below 
``` 
mockMvc.perform(get("/studies?accessionVersionId=EGAS0001")).andExpect(status().is4xxClientError())
                .andExpect(jsonPath("$" +
                ".message").value("Please provide an ID in the form accession.version")); 
``` . If we don't need these we can remove it as i guess.

File: src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/StudyServiceImpl.java (hunk 2 above)
Comment (selvaebi) [reply src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/StudyServiceImpl.java]: This method is used to bypass Spring AOP's methods that returns  released studies or non-deprecated only .

File: src/main/java/uk/ac/ebi/ampt2d/metadata/exceptionhandling/ExceptionHandlers.java [hunk 4]
```diff
@@ -33,9 +34,18 @@
         return new ResponseEntity(new ErrorMessage(HttpStatus.BAD_REQUEST, ex, ex.getMessage()), HttpStatus.BAD_REQUEST);
     }
 
+    @ExceptionHandler(value = ConstraintViolationException.class)
+    public ResponseEntity<ErrorMessage> handleConstraintViolationException(ConstraintViolationException ex) {
+        if (ex.getErrorCode() == 23505) { // 23505 is a constraint violation error Code for duplicate id or index
+            return new ResponseEntity(new ErrorMessage(HttpStatus.CONFLICT, ex,
+                    "Duplicate id or index provided"), HttpStatus.CONFLICT);
```
cyenyxe: Why not using the message from the exception? This message doesn't specify the ID or index so it is not very useful.

src/main/java/uk/ac/ebi/ampt2d/metadata/exceptionhandling/ExceptionHandlers.java
```diff
@@ -33,9 +34,18 @@
         return new ResponseEntity(new ErrorMessage(HttpStatus.BAD_REQUEST, ex, ex.getMessage()), HttpStatus.BAD_REQUEST);
     }
 
+    @ExceptionHandler(value = ConstraintViolationException.class)
+    public ResponseEntity<ErrorMessage> handleConstraintViolationException(ConstraintViolationException ex) {
+        if (ex.getErrorCode() == 23505) { // 23505 is a constraint violation error Code for duplicate id or index
```
cyenyxe: 😨 Don't they provide a nicer way to check this?

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Taxonomy.java
cyenyxe: Given that we are using 'id' everywhere else, I would go for 'taxonomyId'. I would prefer not to shorten other words unless absolutely necessary.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Analysis.java
```diff
@@ -122,7 +134,12 @@
     @ManyToMany
     private List<File> files;
 
-    public AccessionVersionEntityId getId() {
+    public AccessionVersionId getAccessionVersionId() {
+        return accessionVersionId;
+    }
+
+    @Override
+    public Long getId() {
```
cyenyxe: It is a minor detail, but maybe it would make more sense to list `getId` first in all the models, given that is the main identifier instead of accession+version.

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/WebResource.java
```diff
@@ -60,7 +63,7 @@
     @ApiModelProperty(position = 3, required = true)
     @NotNull
     @JsonProperty
-    @Column(nullable = false, columnDefinition = "TEXT")
+    @Column(nullable = false, columnDefinition = "VARCHAR")
```
cyenyxe: Why do we need to set the type here, specially if it's going to be VARCHAR (which I suppose is the default).

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/idconverter/CustomBackendIdConverter.java
cyenyxe: Yes, we still need queries by accession+version.

src/main/java/uk/ac/ebi/ampt2d/metadata/rest/controllers/StudyRestController.java
```diff
@@ -157,10 +156,10 @@
     }
 
     @ApiOperation(value = "Get a list of studies linked to a given study")
-    @ApiParam(name = "id", value = "Study's id (accession.version)", type = "string", required = true, example = "EGAS0001.1")
+    @ApiParam(name = "id", value = "StudyId", type = "string", required = true)
     @RequestMapping(method = RequestMethod.GET, path = "{id}/linkedStudies", produces = "application/json")
     @ResponseBody
-    public ResponseEntity<Resources<StudyResource>> getLinkedStudies(@PathVariable("id") AccessionVersionEntityId id) {
```
cyenyxe: Please do not remove support for queries by accession+version, which is the public identifier for an object.

src/main/java/uk/ac/ebi/ampt2d/metadata/exceptionhandling/ExceptionHandlers.java
selvaebi: As the exception is returned from database ( internal server error , we are changing to client error ) which returns too much information like index id and stuffs which is too much for the user i guess.

cyenyxe: Could you paste an example of the returned exception (either here or in JIRA) so we can decide what to display exactly?

selvaebi:
``` {
  "cause": {
    "cause": {
      "cause": null,
      "message": "Unique index or primary key violation: \"UK_OLRBTSBQD21KAF4CP7RH4RCHN_INDEX_1 ON PUBLIC.TAXONOMY(TAXONOMY_IDENTIFIER) VALUES (1, 1)\"; SQL statement:\ninsert into taxonomy (id, last_modified_date, name, taxonomy_identifier) values (null, ?, ?, ?) [23505-196]"
    },
    "message": "could not execute statement"
  },
  "message": "could not execute statement; SQL [n/a]; constraint [\"UK_OLRBTSBQD21KAF4CP7RH4RCHN_INDEX_1 ON PUBLIC.TAXONOMY(TAXONOMY_IDENTIFIER) VALUES (1, 1)\"; SQL statement:\ninsert into taxonomy (id, last_modified_date, name, taxonomy_identifier) values (null, ?, ?, ?) [23505-196]]; nested exception is org.hibernate.exception.ConstraintViolationException: could not execute statement"
} ```

File: src/main/java/uk/ac/ebi/ampt2d/metadata/exceptionhandling/ExceptionHandlers.java (hunk 5 above)
Comment (selvaebi) [reply src/main/java/uk/ac/ebi/ampt2d/metadata/exceptionhandling/ExceptionHandlers.java]: nope all the constraint violations of the database will be returned by this exception only, it can't be differentiated other than error codes i guess. However all other constraint checks are checked before hand by javaxValidations. Despite that just for sure duplicate key constraint violation i added this checking.

File: src/main/java/uk/ac/ebi/ampt2d/metadata/rest/controllers/StudyRestController.java (hunk 8 above)
Comment (selvaebi) [reply src/main/java/uk/ac/ebi/ampt2d/metadata/rest/controllers/StudyRestController.java]: after discussing with Jorge , any object can be created without an accession , so i think accession cannot be used as an identifier , we can have a queryByAccession but not as an identifier . And also i need to define uniqueConstraints for the objects to stop updates using Post. So as accessions can be null i can't use it as a unique constraint now .

Comment (cyenyxe) [reply src/main/java/uk/ac/ebi/ampt2d/metadata/rest/controllers/StudyRestController.java]: Yeah well, call it "public alias" instead of "public identifier" if you prefer to do so.

> So as accessions can be null i can't use it as a unique constraint now .

Read the last paragraph from [this section](https://www.postgresql.org/docs/current/static/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS). The SQL standard supports unique constraints in nullable columns.

Comment (selvaebi) [reply src/main/java/uk/ac/ebi/ampt2d/metadata/rest/controllers/StudyRestController.java]: Yes but then , the goal of the jira stop update using post method can't be achieved i guess , same post will create new object (not even merge).

File: src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/StudyServiceImpl.java [hunk 9]
```diff
@@ -112,24 +119,25 @@ public int compare(Study o1, Study o2) {
     }
 
     @Override
-    public List<Study> findLinkedStudies(AccessionVersionEntityId id) {
-        Study study = studyRepository.findOne(id);
-        if ( study == null ) {
+    public List<Study> findLinkedStudies(AccessionVersionId accessionVersionId) {
+        Study study = findOneStudyByAccessionVersionId(accessionVersionId);
+        //findOne returns only released/non-deprecated study
```
iyangar: You may want to expand/format the comment as necessary

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/StudyServiceImpl.java
```diff
@@ -112,24 +119,25 @@ public int compare(Study o1, Study o2) {
...
+        if (study == null && studyRepository.findOne(study.getId())==null) {
```
iyangar: Indentation with `==`

Zhicheng-Liu:
As discussed in yesterday's dev meeting, could we make sure that every object that has accession version id provides:
1). search by case insensitive accession through QueryDSL;
2). get the lastest version of the object based on a given accession (through `/accession` endpoint)
3). test cases to cover all these?

src/test/java/uk/ac/ebi/ampt2d/metadata/MetadataApplicationTest.java
```diff
@@ -765,6 +770,20 @@ public void testAccessionValidation() throws Exception {
...
+        postTestStudy("EGAS0001", 1, "test_study", taxonomyUrl);
+        postTestFile("EGAF0001", 1);
+        postTestSample("EGAN0001", "Sample1");
+        mockMvc.perform(get("/studies?accessionVersionId=EGAS0001.1")).andExpect(status().isOk())
+                .andExpect(jsonPath("$..studies[0].accessionVersionId.accession").value("EGAS0001"));
+        mockMvc.perform(get("/files?accessionVersionId=EGAF0001.1")).andExpect(status().isOk())
+                .andExpect(jsonPath("$..studies[0].accessionVersionId.accession").value("EGAF0001"));
+        mockMvc.perform(get("/samples?accessionVersionId=EGAN0001.1")).andExpect(status().isOk())
+                .andExpect(jsonPath("$..studies[0].accessionVersionId.accession").value("EGAN0001"));
+    }
```
iyangar: A search by non-existing AccessionVersionId can be added to make it 360<sup>o</sup>; Optional

selvaebi: ya it is tested in testAccessionValidation .

src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/AccessionVersionId.java
```diff
@@ -20,19 +20,13 @@
...
 import java.io.Serializable;
 
 @Embeddable
 public class AccessionVersionId implements Serializable {
 
-    @Size(min = 1, max = 255)
-    @NotNull
     private String accession;
 
-    @Min(1)
```
cyenyxe: Since we are not going to control these at serialization time, could we do it at construction time, throwing IllegalArgumentException if the values don't fall in the acceptable ranges? One thing is having a null object, and the other constructing one with incorrect values.

src/main/java/uk/ac/ebi/ampt2d/metadata/rest/controllers/StudyRestController.java
```diff
@@ -171,14 +174,14 @@
             "deprecated, it is not possible to update that study object through PATCH /studies/{id} method, it will " +
             "result in NOT FOUND. This new method (PATCH /studies/{id}/patch) allows that study object to be found and" +
             " updated.")
-    @ApiParam(name = "id", value = "StudyId", type = "string", required = true)
+    @ApiParam(name = "id", value = "Study's id", type = "string", required = true, example = "EGAS0001.1")
```
cyenyxe: If this is the patch by ID, then the example is not right.

selvaebi: I removed it already in the latest code.

PR #41, opened 2018-11-20
T2D-269 update metadata dev url and point prod and fallback url to same db

selvaebi:


Jenkinsfile
```diff
@@ -6,7 +6,7 @@ pipeline {
   }
   environment {
     stagingPostgresDbUrl = credentials('STAGINGMETADATAWSDBURL')
-    fallBackPostgresDbUrl = credentials('FALLBACKMETADATAWSDBURL')
+    fallBackPostgresDbUrl = credentials('PRODMETADATAWSDBURL')
```
Zhicheng-Liu: I think we need a discussion on how we want to deploy the DBs.

Jenkinsfile
```diff
@@ -55,7 +55,7 @@ pipeline {
       steps {
         echo 'Deploying to Staging'
         sh "curl --upload-file staging/target/metadata-ws-*.war \
-        'http://'${tomcatCredentials}'@'${stagingHost}':8080/manager/text/deploy?path=/metadata&update=true' | \
+        'http://'${tomcatCredentials}'@'${stagingHost}':8080/manager/text/deploy?path=/dev/metadata&update=true' | \
```
Zhicheng-Liu: Shouldn't we in proxy settings map `/dev/metadata` to `http://staging:8080/metadata`?

selvaebi: the Swagger base path takes host + prefix + servletContext only. So the if we deploy using /metadata - the staging app base url forms to ebi.ac.uk+ ega/ampt2d + metadata - which is again prod app base url.

Zhicheng-Liu: Ok! Then I am fine with this change.

Jenkinsfile
selvaebi: So until we get a production postgres and proxy corrected we gonna use same database for fallback and prod

Zhicheng-Liu: OK! Once we have production instances set up and properly configured for cross instance replication we can use different databases for fallback and production. For now, we have to use the same database.

metadata-ws/Jenkinsfile
```diff
@@ -6,7 +6,7 @@ pipeline {
   }
   environment {
     stagingPostgresDbUrl = credentials('STAGINGMETADATAWSDBURL')
-    fallBackPostgresDbUrl = credentials('FALLBACKMETADATAWSDBURL')
+    fallBackPostgresDbUrl = credentials('PRODMETADATAWSDBURL')
```
iyangar: Is this because we don't have dedicated server at this moment.

PR #40, opened 2018-11-13
T2D-258 file object converter

selvaebi:

PR #42, opened 2018-11-26
T2D-259 Find File object in ENA database

iyangar:
Pulls all ENA XMLs from DB
Parses XML file and extracts the files into a list of AnalysisFileType objects


cyenyxe:
The code is not rebased against the very latest in #40, because it uses metadata-api as module name when it should be metadata-ws.

iyangar:
Rebased

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSet.java
```diff
@@ -0,0 +1,157 @@
...
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.SQLXML;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+
+public class AnalysisFileTypeFromSet implements TypeFromSet<AnalysisFileType, AnalysisSetType> {
```
selvaebi: blank line after class header.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSet.java
```diff
@@ -0,0 +1,157 @@
...
+        return properties;
+    }
+
+    private String createUrl(Properties appProperties) {
+        String url = appProperties.getProperty("OracleJDBC.url");
+        String host = appProperties.getProperty("OracleJDBC.host");
+        String protocol = appProperties.getProperty("OracleJDBC.protocol");
+        String serviceName = appProperties.getProperty("OracleJDBC.servicename");
+        String port = appProperties.getProperty("OracleJDBC.port");
+        url = url + ":@(DESCRIPTION =(ADDRESS_LIST =(ADDRESS = (PROTOCOL = " + protocol + ")(HOST = " + host + ")(PORT = " +
```
selvaebi: Url can be made in the properties file i guess

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSet.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public List<AnalysisFileType> extractFromSqlXml(SQLXML sqlXml) {
+        String xmlStr;
+        List<AnalysisFileType> analysisFileList = new ArrayList<>();
+
+        try {
+            xmlStr = sqlXml.getString();
+            AnalysisSetType analysisSet;
+            try {
```
selvaebi: we can have multiple catches for same try block , no need to nest try blocks .

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSet.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public List<AnalysisFileType> getAnalysisFileObject() throws IOException, SQLException {
+        String prepSql = "select ANALYSIS_XML from ERA.ANALYSIS";
+        Connection dbConnection = null;
+        PreparedStatement prepStatement = null;
+        ResultSet resultSet = null;
+        SQLXML sqlXml;
+        List<AnalysisFileType> analysisFileList = new ArrayList<>();
+
+        Properties appProperties = loadPropertiesFile();
```
selvaebi: we can make use of seperate property bean .

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSetTest.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@RunWith(MockitoJUnitRunner.class)
+public class AnalysisFileTypeFromSetTest {
+    @Test
+    public void testAnalysisFileParser() throws Exception {
+        String xmlStr = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
```
selvaebi: we can read xml file from resources , if we want .

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSetTest.java
```diff
@@ -0,0 +1,95 @@
...
+    public void testAnalysisFileParserWrongInput() throws Exception {
+        String xmlStr = "/<?wrong xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
+                "<ANALYSIS_SET>\n" +
+                "  <ANALYSIS alias=\"uk10k_scoop5013826.vcf.gz-vcf_analysis-sc-20120330\" center_name=\"SC\" broker_name=\"EGA\" analysis_center=\"SC\" accession=\"ERZ000011\">\n" +
+                "  </ANALYSIS>\n" +
+                "</ANALYSIS_SET>\n";
+
+        AnalysisFileTypeFromSet analysisFileTypeFromSet = new AnalysisFileTypeFromSet();
+        Throwable exception = assertThrows(XmlException.class, () -> analysisFileTypeFromSet.getAnalysisSet(xmlStr));
+        assertEquals("error: Unexpected element: CDATA", exception.getMessage());
```
selvaebi: i guess we need to throw/log exceptions with analysis id to refer to the cause of failure

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSet.java
iyangar: Well, i gave options on all parameters in properties file. Let me know, i can make full URL there

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSet.java
```diff
@@ -0,0 +1,157 @@
...
+        String protocol = appProperties.getProperty("OracleJDBC.protocol");
+        String serviceName = appProperties.getProperty("OracleJDBC.servicename");
+        String port = appProperties.getProperty("OracleJDBC.port");
+        url = url + ":@(DESCRIPTION =(ADDRESS_LIST =(ADDRESS = (PROTOCOL = " + protocol + ")(HOST = " + host + ")(PORT = " +
+                port + ")))(CONNECT_DATA =(SERVER = DEDICATED)(SERVICE_NAME = " + serviceName + ")))";
+
+        return url;
+    }
+
+    public List<AnalysisFileType> getAnalysisFileObject() throws IOException, SQLException {
```
Zhicheng-Liu: The class name is `AnalysisFileTypeFromSet` and it implements `TypeFromSet` interface, while this method is about from connecting to database all the way through to getting the list of `AnalysisFileType` objects. I think this class has too many more responsibilities than its name suggests and a lot of logics here are not re-usable for other object types where they should be. Please try to abstract those common logics to re-usable components and re-design the interfaces and their implementations.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/TypeFromSet.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package uk.ac.ebi.ampt2d.metadata.parser;
+
+import java.util.List;
+
+public interface TypeFromSet<SUBSET, SUPERSET> {
```
Zhicheng-Liu: It is definitely not superset - subset relationship. It is more like a collection vs element to me.

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSetTest.java
Zhicheng-Liu: +1

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSet.java
Zhicheng-Liu: +1

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSet.java
iyangar: done

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSet.java
iyangar: Done

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSetTest.java
iyangar: Done

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromSetTest.java
iyangar: done

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/TypeFromSet.java
iyangar: done

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/database/JdbcOperation.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(JdbcOperation.class);
+
+    private JdbcConnection jdbcConnection;
+
+    public JdbcOperation(JdbcConnection jdbcConnection) {
+        this.jdbcConnection = jdbcConnection;
+    }
+
+    private String createCompleteUrl(JdbcConnection jdbcConnection) {
```
cyenyxe: Since we are already using Spring, why not letting it deal with the database connection? This kind of _ad hoc_ JDBC code always fails to support some corner case.

metadata-load/src/main/resources/application.properties
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+#JDBC properties entry for Oracle server
+OracleJDBC.url=jdbc:oracle:thin
+OracleJDBC.host=ora-vm-069.ebi.ac.uk
```
cyenyxe: Do not store hostnames or passwords in the properties file.

iyangar: I will remove hostname also

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/database/JdbcOperation.java
iyangar: To my knowledge JDBC was suggested by ENA team. I can use JDBCTemplate method

jorizci: Just to save you some time...

https://github.com/EbiEga/ega-production/blob/master/file-re-encrypt/src/main/java/uk/ac/ebi/ega/file/re/encrypt/DatabaseConfiguration.java

That is how you configure to read the connection details from properties and create an appropriate DataSource, TransactionManager, JDBCTemplate and if there is more than one connection in the application  also configure properly the HikariCP.

iyangar: done

metadata-load/src/main/resources/application.properties
iyangar: done

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/enaobject/EnaObjectCollect.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public List<AnalysisFileType> getEnaAnalysisFileFromDb(DataSource dataSource) {
+        List<SQLXML> sqlxmlList;
+        String sqlAnalysis = "SELECT ANALYSIS_XML FROM ERA.ANALYSIS";
+        String columnAnalysisXml = "ANALYSIS_XML";
+
+        SqlxmlJdbcTemplate sqlxmlJdbcTemplate = new SqlxmlJdbcTemplate(dataSource,
+                sqlAnalysis, columnAnalysisXml);
+        sqlxmlList = sqlxmlJdbcTemplate.listSqlxml();
```
Zhicheng-Liu: I still think the above process of creating sql statement, fetching data from database should be wrapped behind a data service interface. At least for now we have two sources of data, one directly from the database and the other through REST API. Isn't it better to get the XML through an interface here?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/TypeFromXml.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ *
+ */
+package uk.ac.ebi.ampt2d.metadata.parser;
+
+import java.util.List;
+
+public interface TypeFromXml<ELEMENT, XML> {
+
+    List<ELEMENT> extractFromSqlXml(XML str);
```
Zhicheng-Liu: I believe this method is also intended for extracting from XMLs from other sources. If that is the case could you not name the method name with `SqlXml` in it? It is misleading to me.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/service/DbService.java
```diff
@@ -0,0 +1,38 @@
...
+
+import uk.ac.ebi.ampt2d.metadata.database.SqlxmlJdbcTemplate;
+
+import javax.sql.DataSource;
+import java.sql.SQLXML;
+import java.util.List;
+
+public class DbService {
+
+    private String sqlAnalysis = "SELECT ANALYSIS_XML FROM ERA.ANALYSIS";
```
Zhicheng-Liu: `static final` please.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/database/SqlxmlJdbcTemplate.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ */
+package uk.ac.ebi.ampt2d.metadata.database;
+
+import org.springframework.jdbc.core.JdbcTemplate;
+
+import java.sql.SQLXML;
+import java.util.List;
+
+public class SqlxmlJdbcTemplate {
```
selvaebi: SqlXmlJdbcTemplate ?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/enaobject/EnaObjectCollect.java
```diff
@@ -0,0 +1,59 @@
...
+import uk.ac.ebi.ampt2d.metadata.service.EnaDbService;
+import uk.ac.ebi.ena.sra.xml.AnalysisFileType;
+
+import java.sql.SQLException;
+import java.sql.SQLXML;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class EnaObjectCollect {
```
selvaebi: EnaObjectCollector

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/database/SqlxmlJdbcTemplate.java
iyangar: If the clarification is on name, we can use this class to get any XML file from other tables apart from Analysis. Hence named it as generic SqlXmlJdbcTemplate

iyangar: Since the package is java.sql.```SQLXML``` i took it as single word. We could change to SqlXml aswell. Let me know.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/database/SqlxmlJdbcTemplate.java
```diff
@@ -0,0 +1,43 @@
...
+    private String sql;
+    private String column;
+
+    public SqlxmlJdbcTemplate(JdbcTemplate jdbcTemplate, String sql, String column) {
+        this.jdbcTemplate = jdbcTemplate;
+        this.sql = sql;
+        this.column = column;
+    }
+
+    public List<SQLXML> listSqlxml() {
```
selvaebi: change method name to getSqlXmlList ?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/enaobject/EnaObjectCollector.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class EnaObjectCollector {
+
+    AnalysisFileTypeFromXml analysisFileTypeFromXml = new AnalysisFileTypeFromXml();
+    private static final Logger logger = LoggerFactory.getLogger(EnaObjectCollector.class);
+
+    private List<AnalysisFileType> getAnalysisList(SQLXML sqlxml) {
```
selvaebi: method name should be getAnalysisFileTypeList

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromXml.java
```diff
@@ -0,0 +1,74 @@
...
+            analysisSetType = ANALYSISSETDocument.Factory.parse(xmlStr).getANALYSISSET();
+        } catch (XmlException e) {
+            logger.error("Unable to parse the XML file: {}",  xmlStr, e);
+            throw e;
+        }
+        return analysisSetType;
+    }
+
+    @Override
+    public List<AnalysisFileType> extractFromXml(String xmlStr) {
```
selvaebi: change method name to extractAnalysisFileTypeFromXml

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromXml.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.List;
+
+public class AnalysisFileTypeFromXml implements TypeFromXml<AnalysisFileType, String> {
+
+    private static final Logger logger = LoggerFactory.getLogger(AnalysisFileTypeFromXml.class);
+
+    private List<AnalysisFileType> extract(AnalysisSetType analysisSet) {
+        List<AnalysisFileType> subAnalysisFileList = new ArrayList<>();
+
+        AnalysisType[] analysisType = analysisSet.getANALYSISArray();
```
selvaebi: change the variable name to plural

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromXml.java
```diff
@@ -0,0 +1,74 @@
...
+        for (AnalysisType analysis : analysisType) {
+            AnalysisType.FILES analysisFILES = analysis.getFILES();
+            AnalysisFileType[] analysisFile = analysisFILES.getFILEArray();
```
selvaebi: change the variable name to plural or add Array

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromXml.java
```diff
@@ -0,0 +1,74 @@
...
+            for (AnalysisFileType anaFile : analysisFile) {
```
selvaebi: please change anaFile name

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromXml.java
iyangar: Since its interface implementation extractFromXml is used. To extract different type of objects from XML file.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/service/EnaDbService.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.springframework.jdbc.core.JdbcTemplate;
+import uk.ac.ebi.ampt2d.metadata.database.SqlxmlJdbcTemplate;
+
+import java.sql.SQLXML;
+import java.util.List;
+
+public class EnaDbService {
+
+    private static final String sqlAnalysis = "SELECT ANALYSIS_XML FROM ERA.ANALYSIS";
```
selvaebi: static final variable names we use should be all caps and words separated by an underscore.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/database/SqlxmlJdbcTemplate.java
iyangar: As we discussed changing to getSqlxmlList

iyangar:
Since other classes have DB connectivity dependency i couldn't add test cases for them. I will add failure test case

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/enaobject/EnaObjectCollector.java
```diff
@@ -0,0 +1,62 @@
...
+import java.sql.SQLException;
+import java.sql.SQLXML;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class EnaObjectCollector {
+
+    AnalysisFileTypeFromXml analysisFileTypeFromXml = new AnalysisFileTypeFromXml();
+    private static final Logger logger = LoggerFactory.getLogger(EnaObjectCollector.class);
```
selvaebi: final static variable should be in capitals.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/enaobject/EnaObjectCollector.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private List<AnalysisFileType> getAnalysisFileTypeList(SQLXML sqlxml) {
```
selvaebi: private functions go below public functions

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/enaobject/EnaObjectCollector.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.sql.SQLException;
+import java.sql.SQLXML;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class EnaObjectCollector {
+
+    AnalysisFileTypeFromXml analysisFileTypeFromXml = new AnalysisFileTypeFromXml();
```
selvaebi: use as a local variable or as a component

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/parser/AnalysisFileTypeFromXml.java
```diff
@@ -0,0 +1,68 @@
...
+import uk.ac.ebi.ena.sra.xml.AnalysisType;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class AnalysisFileTypeFromXml implements TypeFromXml<AnalysisFileType, String> {
+
+    private static final Logger logger = LoggerFactory.getLogger(AnalysisFileTypeFromXml.class);
+
+    private List<AnalysisFileType> extract(AnalysisSetType analysisSet) {
```
selvaebi: change method name to extractAnalysisFilesFromAnalysis

metadata-load/pom.xml
```diff
@@ -30,11 +30,81 @@
...
+                    <version>12.2.0.1</version>
+                </dependency>
+                <dependency>
+                    <groupId>com.oracle.jdbc</groupId>
+                    <artifactId>xdb6</artifactId>
+                    <version>12.2.0.1</version>
+                </dependency>
+                <dependency>
+                    <groupId>com.oracle.xml</groupId>
+                    <artifactId>xmlparserv2</artifactId>
```
jorizci: Xmlparserv2 doesn't seem to be needed anymore on version 12.2.0.1, plus it is impossible to download that version from the oracle maven repository, I remember reading a while ago that it was only meant for retrocompatibility. As far as I am aware the tests still pass without it.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/database/DatabaseConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+public class DatabaseConfiguration {
+
+    @Bean ("db_datasource_properties")
+    @ConfigurationProperties("ena.datasource.db")
+    public DataSourceProperties dbDataSourceProperties() {
+        return new DataSourceProperties();
+    }
+
+    @Bean ("db_datasource")
+    @ConfigurationProperties("ena.datasource.db.hikari")
```
jorizci: This won't work, I've checked the libraries that are pulled with this release train. Using **this version of spring** you don't have a hikari connection pool, so that bit is pointless, as Spring will use the Apache connection pool. This is something they dropped on newer versions and exchanged for Hikari. So either we move up the spring version or we modify this and the properties to configure properly the apache connection pool.

jorizci:
On another note, can we fix the problem in the pom?
This line on the metadata-ws pom `<directory>${buildDirectory}</directory>` needs to be removed
And `<outputDirectory>${buildDirectory}/generated-sources/java</outputDirectory>` needs to be changed to `<outputDirectory>target/generated-sources/java</outputDirectory>`

Zhicheng-Liu:
The `${buildDirectory}` variable is currently being used in Jenkins pipeline, so don't remove it. It needs a proper fix.

metadata-load/pom.xml
iyangar: When i remove it i am getting java.lang.NoClassDefFoundError: oracle/xml/binxml/BinXMLMetadataProvider; will explore and discuss with you

iyangar:
https://github.com/EBIvariation/metadata-ws/pull/42#issuecomment-454717945
A new ticket https://www.ebi.ac.uk/panda/jira/browse/T2D-279 has been created as this is a separate task

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/database/DatabaseConfiguration.java
iyangar: Team discussed in slack and a decision on moving to new Spring version is to be made in development meeting

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/service/EnaDbService.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.stereotype.Service;
+
+import java.sql.SQLXML;
+import java.util.List;
+import java.util.Map;
+
+@Service
+public class EnaDbService {
+
+    private static final String SQL_ANALYSIS = "SELECT ANALYSIS_XML FROM ERA.ANALYSIS WHERE ROWNUM <= :rowNum";
```
selvaebi: you can put SQL_ANALYSIS query in the properties file so that we can change it in the test.

iyangar: Can we expose EBI resources. Number of rows will be retrieved from properties file. Is this flexibility enough? let me know

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/enaobject/EnaObjectCollector.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class EnaObjectCollector {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(EnaObjectCollector.class);
+
+    @Autowired
+    EnaDbService enaDbService;
```
selvaebi: add private

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/parser/EnaObjectCollectorTest.java
```diff
@@ -0,0 +1,108 @@
...
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+@TestPropertySource("classpath:application.properties")
+
+public class EnaObjectCollectorTest {
+
+    @Mock
+    NamedParameterJdbcTemplate jdbcTemplate;
```
selvaebi: add private to instance variables

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/enaobject/EnaObjectCollector.java
```diff
@@ -0,0 +1,101 @@
...
+    private long totalPages;
+
+    @Autowired
+    EnaDbService enaDbService;
+
+    public void initialize() {
+        pageFrom = pageSize = totalPages = 1;
+    }
+
+    public List<AnalysisFileType> getEnaAnalysisFileTypeFromDb() {
```
cyenyxe: Change this method interface to return a single page, check preconditions such as valid page number or page size, and implement tests to check these different scenarios.

.travis.yml
```diff
@@ -4,3 +4,9 @@ language: java
 
 jdk:
   - oraclejdk8
+
+install:
+  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -PTravis
```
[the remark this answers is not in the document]
iyangar: Travis executes `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V` without profile.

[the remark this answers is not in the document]
cyenyxe: But my question is why do we need install, then script step.

iyangar:
https://github.com/EBIvariation/metadata-ws/pull/42#discussion_r249420261
With Apache Tomcat settings
ena.datasource.db.tomcat.minIdle=2
ena.datasource.db.tomcat.maxActive=2

.travis.yml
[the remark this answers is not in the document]
iyangar: Original plan was to include something like `mvn test -s settings.xml -PTravis` to execute OracleDB tests. Since we dont execute tests depending on Orable jars on Travis it can be removed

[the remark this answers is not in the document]
iyangar: Observed Travis build fails with `mvn test -B`. We need to include script test also

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/parser/EnaObjectCollectorTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Test
+    public void testGetEnaAnalysisFileTypeFromDb() {
+        List<AnalysisFileType> analysisFileTypeList = enaObjectCollector.getEnaAnalysisFileTypeFromDb(1, 2);
+        assertAnalysisFileTypeList(analysisFileTypeList);
+    }
+
+    private void assertAnalysisFileTypeList(List<AnalysisFileType> analysisFileList) {
+        assertEquals(analysisFileList.size(), 2);
+        assertEquals(analysisFileList.get(0).getFilename(), "UK10K_SCOOP5013826.vcf.gz");
```
cyenyxe: Instead of calling 4 times the `get` method for each element, call it once and initialize a variable.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/enaobject/EnaObjectCollector.java
```diff
@@ -0,0 +1,73 @@
...
+
+public class EnaObjectCollector {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(EnaObjectCollector.class);
+
+    @Autowired
+    EnaDbService enaDbService;
+
+    public List<AnalysisFileType> getEnaAnalysisFileTypeFromDb(int pageNumber, int pageSize) {
+        if (pageNumber <= 0 || pageSize <= 0) {
```
cyenyxe: Given that code is usually based on 0-index, a more intuitive interface would check for `pageNumber < 0`. Will require changing line 48 too.

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/parser/EnaObjectCollectorTest.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Test
+    public void testGetEnaAnalysisFileTypeFromDb() {
+        List<AnalysisFileType> analysisFileTypeList = enaObjectCollector.getEnaAnalysisFileTypeFromDb(1, 2);
+        assertAnalysisFileTypeList(analysisFileTypeList);
+    }
+
+    private void assertAnalysisFileTypeList(List<AnalysisFileType> analysisFileList) {
+        assertEquals(analysisFileList.size(), 2);
```
cyenyxe: The expected value should go first, see [Javadoc](http://junit.sourceforge.net/javadoc/org/junit/Assert.html).

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/parser/EnaObjectCollectorDbTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    @Category(OracleDb.class)
+    public void testGetEnaAnalysisFileTypeFromDb() {
+        /*
+         * This test retrieves Analysis Files from DB.
+         * As the DB content prediction is not possible only size is verified.
+         */
+        List<AnalysisFileType> analysisFileTypeList = enaObjectCollector.getEnaAnalysisFileTypeFromDb(1, 10);
+        assertTrue(analysisFileTypeList.size() >= 10);
```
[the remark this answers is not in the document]
iyangar: When page 1 with size 10 is requested it will map to rows 11 thru 20.
ANALYSIS_XML is non-null and with assumption of each analysis_xml will have atleast 1 file. 
Hence total analysisFileTypes retrieved will be >= 10

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/loader/database/SraDatabaseAnalysisRetriever.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Service;
+import uk.ac.ebi.ampt2d.metadata.loader.SraRetrieverByAccession;
+
+import java.sql.SQLException;
+import java.sql.SQLXML;
+
+@Service
```
selvaebi: @Service should be removed.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/loader/xml/SraAnalysisXmlParser.java
```diff
@@ -28,9 +28,12 @@
 
     private static final Logger LOGGER = Logger.getLogger(SraAnalysisXmlParser.class.getName());
 
+    private final String XML_SET_TAGS = "(<ANALYSIS_SET>|</ANALYSIS_SET>)";
+
     @Override
     public AnalysisType parseXml(String xmlString, String accession) throws XmlException {
-        xmlString = removeRootTagsFromXmlString(xmlString);
+        xmlString = removeRootTagsFromXmlString(xmlString); // For API calls
+        xmlString = removeSetTagsFromXmlString(xmlString); // For database queries
```
Zhicheng-Liu: I wonder whether we should remove the root/set tags in the `SraRetrieverByAccession` concrete classes, so that the XML strings returned from `getXml` methods are well-formed XML strings for the corresponding object types.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/loader/xml/SraXmlParser.java
```diff
@@ -15,16 +15,18 @@
...
 import org.apache.xmlbeans.XmlException;
 
 public abstract class SraXmlParser<SRA_OBJECT> {
 
-    String XML_ROOT_TAGS = "(</ROOT>|<ROOT.*display=xml\">)";
+    private final String XML_ROOT_TAGS = "(</ROOT>|<ROOT.*display=xml\">)";
 
     public abstract SRA_OBJECT parseXml(String xmlString, String accession) throws XmlException;
 
+    protected abstract String removeSetTagsFromXmlString(String xmlString);
```
Zhicheng-Liu: Maybe the implementation for this method could also be provided here to keep it consistent with `removeRootTagsFromXmlString` method. But as I mentioned above, the logic for both methods could be moved to the retriever classes.

cyenyxe: The problem is that the set tags are different for each object type :cry: (ANALYSIS_SET, PROJECT_SET, etc). We could use a more generic regex but I suspect that could backfire at some point.

Zhicheng-Liu: If you look at the root element from the API response, it is using regex to remove. I just checked SRA schema and they all start with `OBJECT_SET` except for `DATASET` which is `DATASETS`.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/loader/xml/SraAnalysisXmlParser.java
cyenyxe: I wanted to keep the responsibilities of each class as delimited as possible, in the sense that one would be only for querying and another for parsing XML (including all its considerations). And because it is hard to implement a 100%-guaranteed generic `removeSetTagsFromXmlString` method (see other comment) I'm not sure it will be highly beneficial to move to `SraRetrieverByAccession`.

PR #45, opened 2019-02-12
T2D-259 Extracted XML parsing from ENA API querying

cyenyxe:
In order to share as much code as possible when querying ENA, either via the API or the DB, the logic for parsing the XML has been extracted from that of the API queries.

Some classes have also been made more generic, and can be derived to support more object types.


metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/loader/core/xml/SraAnalysisXmlParser.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.apache.xmlbeans.XmlException;
+import uk.ac.ebi.ampt2d.metadata.loader.api.SraApiAnalysisRetriever;
+import uk.ac.ebi.ena.sra.xml.ANALYSISDocument;
+import uk.ac.ebi.ena.sra.xml.AnalysisType;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class SraAnalysisXmlParser extends SraXmlParser<AnalysisType> {
```
selvaebi: I think we should have SraAnalysisDocumentXmlParser , because the xml from db uses AnalysisSetDocument.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/loader/core/xml/SraAnalysisXmlParser.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(SraApiAnalysisRetriever.class.getName());
```
selvaebi: SraAnalysisXmlParser.class.getName() ?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/loader/core/xml/SraAnalysisXmlParser.java
cyenyxe: I used simply "analysis" to reflect the fact that what we are parsing represents an analysis, not to make it dependent on a particular class name from the ENA API (that could always change).

selvaebi: yes, I agree.

PR #46, opened 2019-02-28
T2D-271 Add Cucumber test for metadata API

Zhicheng-Liu:
Adding part of the implementation for now. @iyangar is going to work on the rest.


metadata-ws/pom.xml
```diff
@@ -14,6 +14,10 @@
         <version>1.0-SNAPSHOT</version>
     </parent>
 
+    <properties>
+        <cucumber.version>2.4.0</cucumber.version>
```
selvaebi: why not use the latest version?

metadata-ws/src/test/resources/features/web-resource.feature
```diff
@@ -0,0 +1,154 @@
...
+
+  Scenario: register a web resource successfully
+    When user request POST /webResources with json data:
+    """
+    {
+      "type": "CENTER_WEB",
+      "resourceUrl": "http://www.ebi.ac.uk"
+    }
+    """
+    And the response code should be 201
```
selvaebi: It should be Then instead of And

metadata-ws/src/test/resources/features/web-resource.feature
```diff
@@ -0,0 +1,154 @@
...
+    And set the URL to TEST_WEB_RESOURCE
+
+    When user request GET with value of TEST_WEB_RESOURCE
+    Then the response code should be 200
+    Then the result should have type with value CENTER_WEB
```
selvaebi: It would be better if we have And for the next two lines

metadata-ws/src/test/resources/features/web-resource.feature
```diff
@@ -0,0 +1,154 @@
...
+
+  Scenario Outline: register a web resource with valid URL should succeed
+    When user request POST /webResources with json data:
+    """
+    {
+      "type": "CENTER_WEB",
+      "resourceUrl": "<url>"
+    }
+    """
+    And the response code should be 201
```
selvaebi: Then

metadata-ws/src/test/resources/features/web-resource.feature
```diff
@@ -0,0 +1,154 @@
...
+
+
+  # Few common tests
+  Scenario: perform options
+    When user request OPTIONS with header Access-Control-Request-Method,Origin value GET,http://www.evil-url.com
+    And the response code should be 200
+    Then the Access-Control-Allow-Origin header should be present with actual value http://www.evil-url.com
+    Then the Allow header should contain GET
+
+  Scenario Outline: verify metadata objects are auditable
```
selvaebi: Common tests should be moved to separate feature

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,467 @@
...
+            throws Exception {
+        String[] urlList = valueKey.split(",");
+        for (int index = 0; index < lenght; index++) {
+            CommonStates.getResultActions().andExpect(jsonPath("$.."+className+"["+index+"].."+field+".href")
+                    .value(CommonStates.getUrl(urlList[index])));
+        }
+    }
+
+    @And("^the href list of the (.*) of (.*) has items (.*) and (.*)$")
+    public void checkResponseLinkedObjectHrefList1(String field, String className, String fieldValue1, String fieldValue2)
```
Zhicheng-Liu: `checkResponseLinkedObjectHref`, `checkResponseLinkedObjectHrefListLoop` and `checkResponseLinkedObjectHrefList1` are doing essentially the same check. Could they be merged into one step? Either checking a single item or checking a collection.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,467 @@
...
+        CommonStates.getResultActions().andExpect(jsonPath("$.."+className+".href")
+                .value(CommonStates.getUrl(valueKey)));
+    }
+
+    @And("^the href list of the (.*) of (.*) (.*) contained in (.*)$")
+    public void checkResponseLinkedObjectHrefListLoop(String field, String className, int lenght, String valueKey)
+            throws Exception {
+        String[] urlList = valueKey.split(",");
+        for (int index = 0; index < lenght; index++) {
+            CommonStates.getResultActions().andExpect(jsonPath("$.."+className+"["+index+"].."+field+".href")
```
Zhicheng-Liu: This is not safe. Please remove the possibilities of out of bound access occuring.

metadata-ws/pom.xml
iyangar: Using 4.2.6 version gives build errors. Is there any compatibility issue?

Zhicheng-Liu: I think so.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,467 @@
...
+    }
+
+    @When("^user request POST (.*) with json data:$")
+    public void performPostOnResourceUriWithJsonData(String resourceUri, String jsonData) throws Exception {
+        CommonStates.setResultActions(mockMvc.perform(post(resourceUri)
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(jsonData.getBytes())));
+    }
+
+    @When("^user request POST with (.*) for Uri (.*) for stringData (.*) for linkedObjectKey and (.*) for linkedObjectClassName")
```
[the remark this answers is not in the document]
iyangar: requires objectMapper

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,467 @@
...
+    }
+
+    @And("^the href of the class (.*) should be (.*)$")
+    public void checkResponseLinkedObjectHref(String className, String valueKey)
+            throws Exception {
+        CommonStates.getResultActions().andExpect(jsonPath("$.."+className+".href")
+                .value(CommonStates.getUrl(valueKey)));
+    }
+
+    @And("^the href list of the (.*) of (.*) (.*) contained in (.*)$")
```
Zhicheng-Liu: Please use `(\\d*)` for regex matching an integer.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,467 @@
...
+        }
+    }
+
+    @When("^user request GET for the (.*) with query param (.*)")
+    public void performGetOnResourcesQuery(String className, String param) throws Exception {
+            CommonStates.setResultActions(mockMvc.perform(get("/" + className + "?" + param)));
+    }
+
+    @When("^user request search for the (.*) with param (.*)")
+    public void performSearchOnResources(String className, String param) throws Exception {
```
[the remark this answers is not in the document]
iyangar: Regarding above 3 comments: Different patterns for search with `?` `search` params `=` are made with different inputs Ex: `"/" + className+param`   `"/" + className + "?" + param`   `"/" + className+"/search/"+param`  Hence different statements for each. To use single method we need to pass strings like `studies/search/xxx` `studies?xxx` `studies/xxx` in the feature file. Is this acceptable in Gherkin English statements?

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
Zhicheng-Liu: I have fixed this in 5aaf388.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
Zhicheng-Liu: This is no longer an issue as it has been removed in 5aaf388.

metadata-ws/pom.xml
Zhicheng-Liu: There are compatibility issues with higher versions for Cucumber and the version of Spring Boot that is currently used. We will address this once we come around to upgrade the Spring Boot version. So, imho, this conversion can be resolved.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/StudySteps.java
```diff
@@ -0,0 +1,138 @@
...
+    public void performGetOnResources(String className) throws Exception {
+        CommonStates.setResultActions(mockMvc.perform(get("/" + className)));
+    }
+
+    @When("^user request GET for (.*) with query param (.*)")
+    public void performGetOnResourcesQuery(String className, String param) throws Exception {
+        CommonStates.setResultActions(mockMvc.perform(get("/" + className + "?" + param)));
+    }
+
+    @When("^user request POST with (.*) for Uri (.*) for stringData (.*) for linkedObjectKey and (.*) for linkedObjectClassName")
```
Zhicheng-Liu: It looks like this step is used only for creating a new `Taxonomy`. Could you make this more specific and plain? Bear in mind that one point of Cucumber test is for non software developers to write the scenarios.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,343 @@
...
+        webResourceRepository.deleteAll();
+    }
+
+    @Before
+    public void cleanStates() {
+        CommonStates.clear();
+    }
+
+    @Before
+    public void setNonExitstingUrls() {
```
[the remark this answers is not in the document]
iyangar: Correct, I feel merging this function into cleanStates is better. 
>Could you only set it when you actually need it?
This is also possible, but setter method to be called in few scenarios and its repetition.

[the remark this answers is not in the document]
Zhicheng-Liu: > This is also possible, but setter method to be called in few scenarios and its repetition.
I could only find two scenarios in which the "non-existing" is used and they are both defined in `study.feature`. There is a `Given` step in Gherkin which is "used to describe the initial context of the system", e.g.:
> Given there is a non-existing study

Zhicheng-Liu:
There is an issue with skipped test because of undefined step. If you run test with `mvn test`,  although it says "build success", there is one skipped test:
>Tests run: 138, Failures: 0, Errors: 0, Skipped: 1
>
And it is caused by an undefined step:
>You can implement missing steps with the snippets below:

When("^user request exhaustive search with dates for the studies base release-date and with the parameters: from= and (\\d+)$", (Integer arg1) -> {
    // Write code here that turns the phrase above into concrete actions
    throw new PendingException();
});
>

Zhicheng-Liu:
@iyangar Please fix the build issue mentioned above.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    @When("^user request OPTIONS with header (.*) value (.*)$")
+    public void performOptionsWithData(String headerList, String valueList) throws Exception {
+        String[] header = headerList.split(",");
+        String[] value = valueList.split(",");
+
+        CommonStates.setResultActions(mockMvc.perform(options("/")
+                .header(header[0], value[0])
+                .header(header[1], value[1])));
```
Zhicheng-Liu: This is not safe. You can't be sure in the future this method is used differently. In fact currently I found this method is used only in one place. Do you really have to parameterise the method? Can you design the step definition differently?

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/StudySteps.java
```diff
@@ -0,0 +1,134 @@
...
+
+@Ignore
+@AutoConfigureMockMvc
+public class StudySteps {
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @Autowired
+    private ObjectMapper objectMapper;
```
Zhicheng-Liu: The `objectMapper` is not used in this class. It can be removed.

metadata-ws/src/test/resources/features/common.feature
```diff
@@ -0,0 +1,100 @@
+Feature: Common object
```
Zhicheng-Liu: I am going to suggest to rename `common.feature` to `misc.feature` or something like that. And separate the functions only used for `misc.feature` to `MiscSteps` class or similar. The original idea of `Common` was for the functions shared by more than one object types.

Also rename this `Feature` to something like `Miscellaneous fuctions` would be more appropriate, imho.

metadata-ws/src/test/resources/features/common.feature
```diff
@@ -0,0 +1,100 @@
...
+
+  Scenario: perform options
```
Zhicheng-Liu: The point of this test case is to test CORS, not about being able to server `OPTIONS` requests.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/StudySteps.java
```diff
@@ -0,0 +1,134 @@
...
+    public void createTestStudy(String testTaxonomyKey) throws Exception {
+        CommonStates.setResultActions(postTestStudy("EGAS0001", 1, "test_human_study", false, LocalDate.now(), testTaxonomyKey));
+    }
+
+    @When("user create a test parameterized study with (.*) for accession, (.*) for version, (.*) for name, (.*) for deprecated, (.*) for releaseDay and (.*) for taxonomy$")
+    public void createTestStudyParameterizedMore(String accession, int version, String name, boolean deprecated, int releaseDay, String testTaxonomyKey) throws Exception {
+        CommonStates.setResultActions(postTestStudy(accession, version, name, deprecated, LocalDate.now().plusDays(releaseDay), testTaxonomyKey));
+    }
+
+    @When("^user request GET for studies")
```
Zhicheng-Liu: I think this can be served by the `performGetOnResourceUri` in `CommonSteps` class. You can just change it to `user request GET for /studies` whereever it is used, I guess.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,349 @@
...
+        CommonStates.setResultActions(mockMvc.perform(post(resourceUri)
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(jsonData.getBytes())));
+    }
+
+    @When("^user request PATCH (.*) with list (.*) for (.*)")
+    public void performPatchOnResourceWithLinkedObject(String urlKey, String linkedObjectUrlKeys,
+                                                       String linkedObjectClassName) throws Exception {
+        List<String> newUrls = null;
+        if (!linkedObjectUrlKeys.equals("NONE")) {
```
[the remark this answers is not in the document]
Zhicheng-Liu: OK, after futher discussion with @iyangar this can not call `CommonStates.getUrls` as it is directly. The difference between them are one is filtering on the null URL with non-existing key while the other is not. The use case for this method is to for the scenario like `update an analysis with invalid reference sequences should fail`. One of considerations there is to update an analysis with a list reference sequences that contains a `null` object.
Since this is the only place where this kind of behaviour is expected, I am happy to leave it as it is.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/StudySteps.java
```diff
@@ -0,0 +1,125 @@
...
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@Ignore
+@AutoConfigureMockMvc
+public class StudySteps {
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @Given("^there is an URL (.*) with key (.*)$")
```
Zhicheng-Liu: Ideally because this step is used in multiple feature files, it should be included in `CommonSteps`.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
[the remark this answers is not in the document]
Zhicheng-Liu: @selvaebi and @cyenyxe, please comment on this when you come to review this pull request.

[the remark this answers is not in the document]
iyangar: After changes, my comment applies to different searches w.r.t study
Redundant GET statements are removed

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/MiscSteps.java
```diff
@@ -0,0 +1,47 @@
...
+
+import static org.junit.Assert.assertTrue;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.options;
+
+public class MiscSteps {
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @When("^user request OPTIONS with headers Access-Control-Request-Method and Origin with values GET and http://www.evil-url.com$")
```
Zhicheng-Liu: `user request OPTIONS / with GET for Access-Control-Request-Method header and http://www.evil-url.com for Origin header`

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/TaxonomySteps.java
```diff
@@ -0,0 +1,52 @@
...
+public class TaxonomySteps {
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @Autowired
+    private ObjectMapper objectMapper;
+
+    @When("^user request POST taxonomies (\\d*) for id, (.*) for name and (.*) for ancestors")
+    public void performPostOnResourceUriWithStringDataAndLink(long id, String name, String linkedObjectUrlKeys) throws Exception {
```
Zhicheng-Liu: Can you rename the method and arguments to better reflect what this method is about?

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,328 @@
...
+        String idStr = linkedObjectUrl.substring(linkedObjectUrl.lastIndexOf('/') + 1);
+        CommonStates.setResultActions(mockMvc.perform(delete(resourceUrl + "/" + className + "/" + idStr)));
+    }
+
+    @When("^user request search for the (.*) with the parameters: (.*)$")
+    public void performSearchOnResourcesWithParameters(String className, String parameters) throws Exception {
+        CommonStates.setResultActions(mockMvc.perform(get("/"+className+"/search?"+parameters)));
+    }
+
+    @When("^user request elaborate find for the (.*) bases (.*) with the parameters: (.*) and (.*)$")
```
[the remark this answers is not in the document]
iyangar: https://github.com/EBIvariation/metadata-ws/pull/46#discussion_r268584965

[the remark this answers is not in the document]
iyangar: >I can only find this step used in study.feature.
Added accession specific .feature; used here also

metadata-ws/src/test/resources/features/accession.feature
```diff
@@ -0,0 +1,68 @@
...
+  Scenario Outline: find different objects by accession and version
+    When I request POST /taxonomies with JSON payload:
+    """
+    {
+      "taxonomyId": 9606,
+      "name": "Homo Sapiens"
+    }
+    """
+    Then set the URL to TEST_TAXONOMY
+    When I create a test parameterized study with EGAS0001 for accession, 1 for version, test_study for name, false for deprecated, 0 for releaseDay and TEST_TAXONOMY for taxonomy
```
selvaebi: can this be done by payload as well

metadata-ws/src/test/resources/features/accession.feature
```diff
@@ -0,0 +1,68 @@
...
+    Then set the URL to TEST_TAXONOMY1
+    When I request POST /taxonomies with JSON payload:
+    """
+    {
+      "taxonomyId": 2,
+      "name": "Species2"
+    }
+    """
+    Then set the URL to TEST_TAXONOMY2
+    When I create a test parameterized sample with EGAN0001 for accession, 1 for version, Sample1 for name and TEST_TAXONOMY1,TEST_TAXONOMY2 for taxonomy
```
selvaebi: You need to define scenarios for each testing like accession study, accession files etc

metadata-ws/src/test/resources/features/accession.feature
```diff
@@ -0,0 +1,68 @@
...
+
+    When I request elaborate find for the studies bases accessionVersionId with the parameters: EGAS0001.2 and =
+    Then the response code should be 200
+    And the result should contain 0 studies
+
+    When I request elaborate find for the studies bases accessionVersionId with the parameters: EGAS0001 and =
+    Then the response code should be 4xx
+    And the result should have message with value Please provide an ID in the form accession.version
+
+    When I request elaborate find for the studies bases accessionVersionId with the parameters: EGAS0001.S1 and =
```
selvaebi: = cannot be sent as a parameter may be operand would be better for reading.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    @When("^user request PATCH (.*) with list (.*) for (.*)")
+    public void performPatchOnResourceWithLinkedObject(String urlKey, String linkedObjectUrlKeys,
+                                                       String linkedObjectClassName) throws Exception {
+        List<String> newUrls = null;
+        if (!linkedObjectUrlKeys.equals("NONE")) {
+            newUrls = Arrays.stream(linkedObjectUrlKeys.split(","))
+                    .map(key -> CommonStates.getUrl(key))
+                    .collect(Collectors.toList());
```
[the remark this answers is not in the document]
iyangar: https://ebi-amp-t2d.slack.com/archives/C9B867XA5/p1553682896006600
Settings appear same as with others

metadata-ws/src/test/resources/features/accession.feature
iyangar: We could, but taxonomy URLs can not be sent directly; need to provide a means to extract from map

metadata-ws/src/test/resources/features/accession.feature
iyangar: ok, then all necessary steps should be repeated for accession study, accession files etc

metadata-ws/src/test/resources/features/accession.feature
```diff
@@ -0,0 +1,68 @@
+Feature: accession object
+
+  Scenario Outline: find different objects by accession and version
```
Zhicheng-Liu: I can see there are mixtures of test cases here: testing the validation of accessionVersionId and testing the searchability of different objects through accessionVersionId. Please separate them. Even though if separated, they may share a lot of steps to creating testing objects first, if you think that is a problem, deal with it in another way. Please don't mix them together.

metadata-ws/src/test/resources/features/accession.feature
```diff
@@ -0,0 +1,68 @@
...
+    """
+    Then set the URL to TEST_TAXONOMY1
+    When I request POST /taxonomies with JSON payload:
+    """
+    {
+      "taxonomyId": 2,
+      "name": "Species2"
+    }
+    """
+    Then set the URL to TEST_TAXONOMY2
```
Zhicheng-Liu: I don't see a reason why you have to create another two taxonomies (`TEST_TAXONOMY1` and `TEST_TAXONOMY2`) here. The taxonomies are not related to the test cases laid out in this scenario. Could you not re-use the other one (`TEST_TAXONOMY`) instead?

metadata-ws/src/test/resources/features/accession.feature
```diff
@@ -0,0 +1,68 @@
...
+    When I create a test parameterized sample with EGAN0001 for accession, 1 for version, Sample1 for name and TEST_TAXONOMY1,TEST_TAXONOMY2 for taxonomy
+    Then the response code should be 201
+    And set the URL to TEST_SAMPLE
+
+    When I request elaborate find for the studies bases accessionVersionId with the parameters: EGAS0001.2 and =
+    Then the response code should be 200
+    And the result should contain 0 studies
+
+    When I request elaborate find for the studies bases accessionVersionId with the parameters: EGAS0001 and =
+    Then the response code should be 4xx
```
Zhicheng-Liu: Ideally the negative test cases should be in a separate scenario. In a separate scenario, negative test cases fail.

metadata-ws/src/test/resources/features/accession.feature
Zhicheng-Liu: I don't like this step either, for its incomprehensibility without digesting the corresponding code. To quote a sentence from Cucumber official site:
> Our open source tool tests **business-readable** specifications against your code on any modern development stack.
The more I looked at it, the more I wanted to change it. And hence 21d294f.

metadata-ws/src/test/resources/features/accession.feature
Zhicheng-Liu: I agree with @selvaebi  here. The secenario here is for testing the searchability of objects through accessionVersionId. That is probabaly one of the most important ways for a client to locate an object. So please make sure that every objects that have `accessionVersionId` can be searched by `accessionVersionId`. That means, in addition to `File`, `Sample` and `Study`, we should also test `Analysis` for this.

Probably there should be step definitions for those four object types which allows creating a default test object or creating a test object with accession and version as arguments. Then you can combine testings on those four different types of object into a "Scenario Outline".

For what it's worth, I don't think the repetition in Cucumber steps is necessarily a bad thing. After all, in Cucumber's context, specifications, tests and documentation are in the **same document**.  They are more like business documents, definitely not code. So we don't treat them in the same way as we treat code in terms of abstraction. I think they should be plain, detailed, easy to understand and repeated wherever needed.

metadata-ws/src/test/resources/features/accession.feature
Zhicheng-Liu: I think one way could make this work is to have placeholders in the payload and on the step handler side, replace the placeholders with real values.

iyangar: I have checked-in payload changes. If everything looks good i can change in all other places

metadata-ws/src/test/resources/features/accession.feature
iyangar: I have separated accession validation and accession search. If any comment still applies, i will address

metadata-ws/src/test/resources/features/accession.feature
iyangar: https://github.com/EBIvariation/metadata-ws/pull/46#discussion_r272830455

metadata-ws/src/test/resources/features/accession.feature
iyangar: https://github.com/EBIvariation/metadata-ws/pull/46#discussion_r272830455.
Separate taxonomy for studies and sample; otherwise one common and other exclusive

metadata-ws/src/test/resources/features/accession.feature
iyangar: https://github.com/EBIvariation/metadata-ws/pull/46#discussion_r272830455

metadata-ws/src/test/resources/features/accession.feature
selvaebi: Creating a scenario for each object(study, taxonomy, analysis,etc.. ) would be clean and presentable.

metadata-ws/src/test/resources/features/analysis.feature
```diff
@@ -0,0 +1,394 @@
+Feature: analysis object
+
+  Scenario: register an analysis successfully
+    When I request POST /reference-sequences with JSON payload:
```
selvaebi: i guess the dependent object creations can be rephrased using "**Given** I had created reference-sequences with JSON payload ... And set url to TEST_REF_SEQ "

metadata-ws/src/test/resources/features/analysis.feature
```diff
@@ -0,0 +1,394 @@
...
+
+    When I create a test analysis with TEST_STUDY for study and TEST_REFERENCE_SEQUENCE_1 for reference sequence
+    Then the response code should be 201
+    And set the URL to TEST_ANALYSIS
+    When I request GET with value of TEST_ANALYSIS
+    Then the response code should be 200
+    And the result should have accessionVersionId.accession with value EGAA0001
+    When I request GET for referenceSequences of TEST_ANALYSIS
+    Then the response code should be 200
+    And the result should contain 1 reference-sequences
```
selvaebi: "contain 1 reference-sequence**s** " is odd

metadata-ws/src/test/resources/features/analysis.feature
iyangar: Ok, in the entire code When-Then is used from beginning. If everybody agrees i can change everywhere.

metadata-ws/src/test/resources/features/analysis.feature
iyangar: `reference-sequences` is class name in json body.

selvaebi: ya you can rephrase the line suitable for singular and plural cases

iyangar: ok, "And the result should contain 1 `reference-sequence` " will be handled separately in reference-sequence specific class

metadata-ws/src/test/resources/features/accession.feature
iyangar: done

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/AnalysisSteps.java
```diff
@@ -0,0 +1,82 @@
...
+    public void createTestAnalysis(String studyUrlKey, String referenceSequenceUrlKey) throws Exception {
+        createTestAnalysisParam("EGAA0001", referenceSequenceUrlKey, studyUrlKey, Analysis.Technology.EXOME_SEQUENCING,
+                Analysis.Type.CASE_CONTROL, "Illumina");
+    }
+
+    @When("I create a test analysis with (.*) for accession, (.*) for reference sequence, (.*) for study, (.*) for technology, (.*) for type and (.*) for platform")
+    public void createTestAnalysisParam(String accession, String referenceSequenceUrlKey, String studyUrlKey,
+                                        Analysis.Technology technology, Analysis.Type type, String platform)
+            throws Exception {
+
```
cyenyxe: Remove blank line.
```suggestion
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/AnalysisSteps.java
```diff
@@ -0,0 +1,82 @@
...
+        CommonStates.setResultActions(
+                postTestAnalysis(accession, referenceSequenceUrlKey, studyUrlKey,
+                        technology, type, platform)
+        );
+    }
+
+    private ResultActions postTestAnalysis(String accession, String referenceSequenceUrlKey, String studyUrlKey,
+                                           Analysis.Technology technology, Analysis.Type type, String platform)
+            throws Exception {
+
```
cyenyxe: Remove blank line.
```suggestion
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,322 @@
...
+    @When("^I request PATCH (.*) with content (.*)")
+    public void performPatchOnResourceWithContent(String urlKey, String content) throws Exception {
+            CommonStates.setResultActions(mockMvc.perform(patch(CommonStates.getUrl(urlKey))
+                    .content(content)));
+    }
+
+    @When("^I request DELETE for the (.*) of (.*) of the (.*)")
+    public void performDeleteOnResourceWithLinkedObject(String className, String linkedObjectUrlKey,
+                                                        String resourceUrlKey) throws Exception {
+
```
cyenyxe:
```suggestion
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    @When("^I request POST (.*) with JSON payload:$")
+    public void performPostOnResourceUriWithJsonData(String resourceUri, String jsonData) throws Exception {
+        CommonStates.setResultActions(mockMvc.perform(post(resourceUri)
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(jsonData.getBytes())));
+    }
+
+    @When("^I request PATCH (.*) with list (.*) for (.*)")
```
cyenyxe: Should `of` be used instead of `for`?

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,322 @@
...
+    public void checkHeaderPresenceWithValue(String headerName, String valueKey) {
+        String headerValue = CommonStates.getUrl(valueKey);
+        assertEquals("Header not present with value: " + headerName + "=" + headerValue,
+                headerValue,
+                CommonStates.getResultActions().andReturn().getResponse().getHeaderValue(headerName));
+    }
+
+    /* check http response body */
+
+    @Then("^the result json should be:$")
```
cyenyxe:
```suggestion
    @Then("^the response JSON should be:$")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,322 @@
...
+    public void checkResponseJsonMatch(String jsonString) throws Exception {
+        CommonStates.getResultActions().andExpect(content().json(jsonString));
+    }
+
+    @Then("^the result should contain (\\d*) (.*)$")
```
cyenyxe:
```suggestion
    @Then("^the response should contain (\\d*) (.*)$")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,322 @@
...
+    public void checkResponseListSize(int size, String className) throws Exception {
+        CommonStates.getResultActions().andExpect(jsonPath("$.."+className).isArray())
+                .andExpect(jsonPath("$.."+className+".length()").value(size));
+    }
+
+    @Then("^the result should have (.*) with value (.*)$")
```
cyenyxe:
```suggestion
    @Then("^the response should contain field (.*) with value (.*)$")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,322 @@
...
+    public void checkResponseJsonFieldValue(String field, String value) throws Exception {
+        CommonStates.getResultActions().andExpect(jsonPath("$."+field).value(value));
+    }
+
+    @Then("^the result should have (.*) as number$")
```
cyenyxe:
```suggestion
    @Then("^the response should contain field (.*) with a numeric value$")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,322 @@
...
+    public void checkResponseJsonFieldValueNumber(String field) throws Exception {
+        CommonStates.getResultActions().andExpect(jsonPath("$."+field).isNumber());
+    }
+
+    @Then("^the result should have (.*) as null$")
```
cyenyxe:
```suggestion
    @Then("^the response should contain field (.*) with null value$")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,322 @@
...
+    public void checkResponseJsonFieldValueNull(String field) throws Exception {
+        CommonStates.getResultActions().andExpect(jsonPath("$."+field).value(nullValue()));
+    }
+
+    @Then("^the result should have (.*) as false$")
```
cyenyxe:
```suggestion
    @Then("^the response should contain field (.*) with a false boolean value$")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,322 @@
...
+        CommonStates.getResultActions().andExpect(jsonPath("$."+field).value(false));
+    }
+
+    @Then("^the difference between (.*) and today should be (\\d*) day$")
+    public void checkResponseJsonFieldValueDay(String field, int diff) throws Exception {
+        LocalDate days = LocalDate.now().plusDays(diff);
+        CommonStates.getResultActions().andExpect(jsonPath("$."+field).value(days.toString()));
+    }
+
+    @Then("^the result should have (.*) existing$")
```
cyenyxe:
```suggestion
    @Then("^the response should contain field (.*)$")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,322 @@
...
+    public void checkResponseJsonFieldValueExist(String field) throws Exception {
+        CommonStates.getResultActions().andExpect(jsonPath("$."+field).exists());
+    }
+
+    @Then("^the result should have (.*) non empty$")
```
cyenyxe:
```suggestion
    @Then("^the response should contain field (.*) with a non-empty value$")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -0,0 +1,322 @@
...
+    public void checkResponseJsonFieldValueNotEmpty(String field) throws Exception {
+        CommonStates.getResultActions().andExpect(jsonPath("$."+field).isNotEmpty());
+    }
+
+    @Then("^the result should not contain (.*)$")
```
cyenyxe:
```suggestion
    @Then("^the response should not contain field (.*)$")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/ReferenceSequenceSteps.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package uk.ac.ebi.ampt2d.metadata.cucumber;
+
+import cucumber.api.java.en.Then;
+
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+
+public class ReferenceSequenceSteps {
+
+    @Then("^the result should contain one reference-sequence$")
```
cyenyxe:
```suggestion
    @Then("^the response should contain one reference sequence$")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/SampleSteps.java
```diff
@@ -0,0 +1,99 @@
...
+        createTestSampleParameterized("EGAS0001", 1, "test_human_sample", testTaxonomyKeys);
+    }
+
+    @When("I create a test parameterized sample with (.*) for accession, (.*) for version, (.*) for name and (.*) for taxonomy")
+    public void createTestSampleParameterized(String accession, int version, String name, String testTaxonomyKeys) throws Exception {
+        List<String> testTaxonomyList = CommonStates.getUrls(testTaxonomyKeys);
+        CommonStates.setResultActions(postTestSample(accession, name, testTaxonomyList, version));
+    }
+
+    @When("I create a test sample no or null accession (.*), (.*) for name and (.*) for taxonomy")
```
cyenyxe: Slightly better written:

```suggestion
    @When("I create a test sample with null or no accession (.*), (.*) for name and (.*) for taxonomy")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/SampleSteps.java
```diff
@@ -0,0 +1,99 @@
...
+@AutoConfigureMockMvc
+public class SampleSteps {
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @Autowired
+    private ObjectMapper objectMapper;
+
+    @When("I create a test sample with (.*) for taxonomy")
```
cyenyxe: The message should not state the objects are 'test'. If written as a user story, nobody will care whether it is a test object or not. Applies to other classes.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/StudySteps.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @When("I create a test study with (.*) for taxonomy$")
+    public void createTestStudy(String testTaxonomyKey) throws Exception {
+        CommonStates.setResultActions(postTestStudy("EGAS0001", 1, "test_human_study", false, LocalDate.now(), testTaxonomyKey));
+    }
+
+    @When("^I request POST /studies with JSONLIKE payload:$")
```
cyenyxe: Why JSONLIKE and not JSON?
```suggestion
    @When("^I request POST /studies with JSON payload:$")
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/StudySteps.java
```diff
@@ -0,0 +1,160 @@
...
+    public void postTestStudy(String jsonLikeData) throws Exception {
+        String[] values = jsonLikeData.split(",");
+        String json = "{";
+        for (String value:values) {
```
cyenyxe: Following the rest of our coding style:
```suggestion
        for (String value : values) {
```

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/SampleSteps.java
iyangar: Would it be better even to rephrase "And set the URL to TEST_SAMPLE" to `And set the URL to SAMPLE`

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/SampleSteps.java
iyangar: To avoid conflict "I create a test sample `having` null or no accession" is used

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/StudySteps.java
iyangar: Most of the content is is like JSON. But taxonomy link is obtained in code

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/SampleSteps.java
iyangar: Changed

metadata-ws/src/test/resources/features/accession.feature
```diff
@@ -0,0 +1,147 @@
+Feature: accession object
```
cyenyxe: Could you please normalize the style in the feature files?

* Some lines use "Scenario" and others "Scenario outline" (the former is more common) 
* Insert one blank line before each "given-when-then" for grouping, right now it seems a bit arbitrary
* Insert 2 blank lines before each "Scenario" clause to represent a greater split than in the "given-when-then"

metadata-ws/src/test/resources/features/sample.feature
```diff
@@ -0,0 +1,340 @@
...
+
+  Scenario Outline: find one sample by taxonomy
+    When I request POST /taxonomies with JSON payload:
+    """
+    {
+      "taxonomyId": 207598,
+      "name": "Homininae"
+    }
+    """
+    Then set the URL to TAXONOMY_1
```
cyenyxe: Why not using valid URLs in all these examples? The system should in fact complain if something like "TAXONOMY_1" is provided.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/CommonSteps.java
```diff
@@ -250,17 +249,17 @@ public void checkResponseJsonFieldValueDay(String field, int diff) throws Except
         CommonStates.getResultActions().andExpect(jsonPath("$."+field).value(days.toString()));
     }
 
-    @Then("^the result should have (.*) existing$")
-    public void checkResponseJsonFieldValueExist(String field) throws Exception {
-        CommonStates.getResultActions().andExpect(jsonPath("$."+field).exists());
+    @Then("^the response should contain field releaseDate$")
```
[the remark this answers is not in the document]
iyangar: Earlier comment https://github.com/EBIvariation/metadata-ws/pull/46#discussion_r267723712 suggested to be specific, when used only for that

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/SampleSteps.java
```diff
@@ -0,0 +1,101 @@
...
+        createTestSampleParameterized("EGAS0001", 1, "test_human_sample", testTaxonomyKeys);
+    }
+
+    @When("I create a parameterized sample with (.*) for accession, (.*) for version, (.*) for name and (.*) for taxonomy")
+    public void createTestSampleParameterized(String accession, int version, String name, String testTaxonomyKeys) throws Exception {
+        List<String> testTaxonomyList = CommonStates.getUrls(testTaxonomyKeys);
+        CommonStates.setResultActions(postTestSample(accession, name, testTaxonomyList, version));
+    }
+
+    @When("I create a no accession sample with (.*) for name and (.*) for taxonomy")
```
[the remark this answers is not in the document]
iyangar: `I create a sample with (.*)` is already present and conflicts. We have to rephrase better.
Like "I create a sample` having`" or "I create a sample `without` accession". Latter works here but not for below comment.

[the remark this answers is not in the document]
cyenyxe: You're right, I didn't consider the regular expression. What about this?

```suggestion
    @When("I create a non-accessioned sample with (.*) for name and (.*) for taxonomy")
```

metadata-ws/src/test/resources/features/accession.feature
iyangar: - Only `Scenario Outline` can have examples section where we can provide parameters for a repetition(loop)
- One blank like after "given-when-then" : done
- 2 blank lines before each "Scenario" : done, had been missed earlier. Corrected everywhere

metadata-ws/src/test/resources/features/sample.feature
iyangar: It was designed to store name with actual URL in map. The same was continued.
Otherwise we need to extract the URL and pass in next statements

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/SampleSteps.java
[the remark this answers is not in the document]
iyangar: Can we clearly interpret `non-accessioned` as `no accession` only and NOT null accession
Below statement becomes "I create a `null-accessioned` sample"
If this looks better i will change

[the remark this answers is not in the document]
cyenyxe: What about these two?

* When I create a non-accessioned sample with (...)
* When I provide a null accession for a sample with (...)

This way there is no problem with the regular expression, and I think they are more readable.

PR #49, opened 2019-03-13
T2D-291 — Store Sample in metadata

selvaebi:

PR #48, opened 2019-03-11
T2D-288 store study in metadata

selvaebi:

PR #47, opened 2019-03-05
T2D-287 store analysis

selvaebi:

PR #50, opened 2019-04-15
T2D-297 — Retrieve reference sequence from ENA

tskir:
### [Diff link](https://github.com/EBIvariation/metadata-ws/pull/50/files)

* Implemented `ReferenceSequenceExtractor` and `ReferenceSequenceConverter classes`.
* Implemented `SraAssemblyXmlParser` class to parse the `ASSEMBLY` documents.
* Modified the application flow to use the new classes.

Points worth noting (& discussing):
* All created reference sequences have type == `ASSEMBLY` because this is the only type which ENA API ever returns
* All created reference sequences have patch == `NOT_SPECIFIED` because ENA data model never specifies the patch directly, only accession+version.


iyangar:
Did you fork the code from https://github.com/EBIvariation/metadata-ws/ to your local branch under `https://github.com/tskir`. It appears like you are directly working on upstream

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/ReferenceSequenceExtractor.java
```diff
@@ -0,0 +1,89 @@
...
+        this.referenceSequenceConverter = referenceSequenceConverter;
+    }
+
+    public ReferenceSequence extractReferenceSequence(String analysisAccession) {
+        ReferenceSequence referenceSequence;
+        try {
+            String analysisXml = sraRetrieverByAccession.getXml(analysisAccession);
+            AnalysisType analysisType = sraAnalysisXmlParser.parseXml(analysisXml, analysisAccession);
+            String assemblyAccession = getAssemblyAccessionOfAnalysis(analysisType);
+            if (assemblyAccession == null) {return null;}  // not all analysis entities will have a reference sequence
```
iyangar: formatting to next line

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/ReferenceSequenceExtractor.java
```diff
@@ -0,0 +1,89 @@
...
+            REFERENCE_SEQUENCE_EXTRACT_SERVICE_LOGGER.log(Level.SEVERE, exception.getMessage());
+        }
+        return null;
+    }
+
+    private String getAssemblyAccessionOfAnalysis(AnalysisType analysis) {
+        String assemblyAccession = null;
+        ReferenceSequenceType referenceSequence = null;
+        AnalysisType.ANALYSISTYPE analysisType = analysis.getANALYSISTYPE();
+        if (analysisType.isSetREFERENCEALIGNMENT()) {referenceSequence = analysisType.getREFERENCEALIGNMENT();}
```
iyangar: same as above

iyangar:
Please add test cases

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/ReferenceSequenceExtractor.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    public ReferenceSequence extractReferenceSequence(String analysisAccession) {
+        ReferenceSequence referenceSequence;
+        try {
+            String analysisXml = sraRetrieverByAccession.getXml(analysisAccession);
+            AnalysisType analysisType = sraAnalysisXmlParser.parseXml(analysisXml, analysisAccession);
+            String assemblyAccession = getAssemblyAccessionOfAnalysis(analysisType);
+            if (assemblyAccession == null) {return null;}  // not all analysis entities will have a reference sequence
+            String assemblyXml = sraRetrieverByAccession.getXml(assemblyAccession);
```
selvaebi: it only works through API , for database connections , in case of ega we need to fgure out what is the query to fetch , and also the accessions can be null in EGA case.

tskir:
> Did you fork the code from https://github.com/EBIvariation/metadata-ws/ to your local branch under https://github.com/tskir. It appears like you are directly working on upstream https://github.com/EBIvariation/metadata-ws/tree/t2d-297-retrieve-reference-sequence

Yeah, it's just that at the time of starting work on this ticket I didn't know that forking is customary. This PR will have to be processed as is because GitHub does not allow to change the base branch, but all subsequent PRs from me will come from the fork.

tskir:
Earlier me and @selvaebi decided on the order of merge for several open PRs in metadata-load. It's going to be:
1. https://github.com/EBIvariation/metadata-ws/pull/48, T2D-288 — store study in metadata
2. https://github.com/EBIvariation/metadata-ws/pull/47, T2D-287 — store sample in metadata
3. https://github.com/EBIvariation/metadata-ws/pull/50, T2D-297 — Retrieve reference sequence from ENA (this PR)
4. https://github.com/EBIvariation/metadata-ws/pull/49, T2D-291 — Store sample in metadata

In line with this, I rebased this PR on top of T2D-287. To view only changes made by me and not by @selvaebi, this link will come in handy for reviewers:
https://github.com/selvaebi/metadata-ws/compare/T2D-287StoreAnalysis...ebivariation:t2d-297-retrieve-reference-sequence

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/ReferenceSequenceExtractor.java
tskir: Fixed: https://github.com/EBIvariation/metadata-ws/pull/50/commits/635dfb1c728c9281ee0de959ba86063fee170df0

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/ReferenceSequenceExtractor.java
tskir: Fixed

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/ReferenceSequenceExtractor.java
tskir: Discussing with @cyenyxe my understanding was that the scope of this ticket is concerned only with retrieving the reference sequence from ENA by API, DB queries will have to be implemented separately

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/ReferenceSequenceExtractor.java
```diff
@@ -0,0 +1,93 @@
...
+            SraXmlParser<AnalysisType> sraAnalysisXmlParser,
+            SraXmlParser<AssemblyType> sraAssemblyXmlParser,
+            Converter<AssemblyType, ReferenceSequence> referenceSequenceConverter) {
+        this.sraRetrieverByAccession = sraRetrieverByAccession;
+        this.sraAnalysisXmlParser = sraAnalysisXmlParser;
+        this.sraAssemblyXmlParser = sraAssemblyXmlParser;
+        this.referenceSequenceConverter = referenceSequenceConverter;
+    }
+
+    public ReferenceSequence extractReferenceSequence(String analysisAccession) {
```
selvaebi: can you please move this code to SraObjectsImporterThroughApi

tskir: Done in https://github.com/EBIvariation/metadata-ws/pull/50/commits/d504b323f7313c1800a3ba0d5872018c12644af0 and https://github.com/EBIvariation/metadata-ws/pull/50/commits/c8d1eb058b4838fbe58325440d5224047036a339

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Analysis.java
```diff
@@ -105,7 +98,7 @@
     @ApiModelProperty(position = 6, dataType = "java.lang.String", example = "[url1, url2]",
             notes = "URL(s) to the reference sequence(s): either a single URL to an assembly/transcriptome, or a comma-separated list of gene sequences")
     @JsonProperty
-    @ManyToMany
+    @ManyToMany(cascade = {CascadeType.PERSIST, CascadeType.MERGE})
```
tskir: As discussed with @selvaebi, this cascading needs to be ditched in favor of storing the reference sequence manually, in order to avoid duplicate issues

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/ReferenceSequence.java
```diff
@@ -87,11 +91,28 @@
     @ManyToOne(optional = false)
     private Taxonomy taxonomy;
```
tskir: As discussed with @selvaebi, taxonomy needs to be retrieved inside ReferenceSequence retrieval as well.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Analysis.java
tskir: Fixed in https://github.com/EBIvariation/metadata-ws/pull/50/commits/767b4d1a20eda626c547e83b812b3ea353078d00

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/ReferenceSequence.java
tskir: Fixed in https://github.com/EBIvariation/metadata-ws/pull/50/commits/767b4d1a20eda626c547e83b812b3ea353078d00

tskir:
@iyangar 

> Please add test cases

Fixed in https://github.com/EBIvariation/metadata-ws/pull/50/commits/43b33e83988d8636aee8559f3cd29714f6666181.

tskir:
Rebased the code on top of the changes recently made by @selvaebi in #48 and #47.

Everything that was mentioned in the reviews is now fixed; **this PR is ready for the next iteration of review.** Merge order & handy diff link for reviewers are in this comment: https://github.com/EBIvariation/metadata-ws/pull/50#issuecomment-487908220

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/ReferenceSequenceExtractor.java
tskir: Discussed this additionally with @selvaebi and @cyenyxe; decided that it doesn't make sense to implement reference sequence retrieval from the database, it will have to be done through API anyway.

selvaebi:
we need to set the study taxonomy to be fetched from the reference sequence as well.

tskir:
> we need to set the study taxonomy to be fetched from the reference sequence as well.

Decided to make this a separate ticket during the sprint meeting because this is not related to the metadata import, but rather to the `metadata-ws` itself

tskir:
It was a bit painful (lots of changes happening!), but I finally rebased this branch on top of #47. It's now ready for the next iteration of review. Diff link: https://github.com/selvaebi/metadata-ws/compare/T2D-287StoreAnalysis...ebivariation:t2d-297-retrieve-reference-sequence

selvaebi:
Hi Kirill, sorry for that but, can you please rebase once again.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/converter/ReferenceSequenceConverter.java
```diff
@@ -0,0 +1,67 @@
...
+        this.referenceSequenceRepository = referenceSequenceRepository;
+    }
+
+    @Override
+    public ReferenceSequence convert(AssemblyType assemblyType) {
+        Taxonomy taxonomy = findExistingorSaveandReturnTaxonomy(assemblyType);
+        return convertSaveAndReturnNewReferenceSequence(assemblyType, taxonomy);
+    }
+
+    private Taxonomy findExistingorSaveandReturnTaxonomy(AssemblyType assemblyType) {
```
selvaebi: Sorry I was wrong before can you please separate converter and persistence ,  like i have done in importStudy's dependentObjects.

metadata-load/src/test/resources/assembly/assemblyDocumentAPI
```diff
@@ -0,0 +1,266 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
selvaebi: Can you rename the xmls and text files in CamelCase and with extension eg: AssemblyDocumentAPI.xml

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/converter/ReferenceSequenceConverter.java
tskir: Fixed in https://github.com/EBIvariation/metadata-ws/pull/50/commits/6ff97f90c3c951061edceeeccc4855a94e78f12a. Now the three processes are separated:

* ReferenceSequence conversion
* Taxonomy extraction, lookup & persistence
* ReferenceSequence lookup & persistence

metadata-load/src/test/resources/assembly/assemblyDocumentAPI
tskir: Fixed in https://github.com/EBIvariation/metadata-ws/pull/50/commits/1847b04057735e88ea0a3fe099dc559c01405dd8.

tskir:
> Hi Kirill, sorry for that but, can you please rebase once again.

Rebased against the latest changes #47. Fixed all other comments. **This request is now ready for the 4th iteration of review.**

[*Handy diff link which includes only changes from this PR*](https://github.com/selvaebi/metadata-ws/compare/T2D-287StoreAnalysis...ebivariation:t2d-297-retrieve-reference-sequence)

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/repositories/ReferenceSequenceRepository.java
```diff
@@ -36,4 +39,6 @@ default void customize(QuerydslBindings bindings, QReferenceSequence referenceSe
                 .first((path, value) -> path.any().equalsIgnoreCase(value.iterator().next()));
     }
 
+    ReferenceSequence findByAccessions(@Param("accessions") List<String> accessions);
```
selvaebi: does this query return only one reference-sequence?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/converter/ReferenceSequenceConverter.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.springframework.core.convert.converter.Converter;
+import uk.ac.ebi.ampt2d.metadata.persistence.entities.ReferenceSequence;
+import uk.ac.ebi.ena.sra.xml.AssemblyType;
+
+import java.util.Arrays;
+
+public class ReferenceSequenceConverter implements Converter<AssemblyType, ReferenceSequence> {
+
+    public ReferenceSequenceConverter() {
```
selvaebi: No need for this constructor

iyangar:
General: metadata-ws module has cucumber test cases. Should tests here be continued as is or converted to cucumber later

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/repositories/ReferenceSequenceRepository.java
tskir: Discussed this during yesterday's standup; decided to leave as is for now. Natural key for the reference key is going to be the list of its accessions. Because of the way it's implemented, reference sequences which have exactly the same list of accessions will not be duplicated, hence this will always return only one reference sequence.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/converter/ReferenceSequenceConverter.java
tskir: Fixed in https://github.com/EBIvariation/metadata-ws/pull/50/commits/1237be8d5509f5705b777378071b4c960f623511

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/MetadataReferenceSequenceFinderOrPersister.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public MetadataReferenceSequenceFinderOrPersister(ReferenceSequenceRepository referenceSequenceRepository) {
+        this.referenceSequenceRepository = referenceSequenceRepository;
+    }
+
+    public ReferenceSequence findOrPersistReferenceSequence(ReferenceSequence referenceSequence) {
+        /* The below find query will make sure to return the same reference sequence when analyses sharing same
+        reference sequence are imported in different runs */
+        ReferenceSequence existingReferenceSequence = referenceSequenceRepository.findByAccessions(
+                referenceSequence.getAccessions());
```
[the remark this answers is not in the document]
tskir: As discussed during the standup yesterday, currently the ReferenceSequence is identified by a full list of its accessions. In the future a more fitting & complicated solution will likely need to be implemented, in a separate ticket. For initial import this solution will work because all identical reference sequences will use the same list of accessions.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/MetadataStudyFinderOrPersister.java
```diff
@@ -0,0 +1,50 @@
...
+    public Study findOrPersistStudy(Study study) {
+        /* The below find query will make sure to return shared study when analyses sharing same study are imported
+          in different runs */
+        Study sharedStudy = studyRepository.findOne(qStudy.accessionVersionId.accession.eq(study
+                .getAccessionVersionId().getAccession()).and(qStudy.accessionVersionId.version.eq(study
+                .getAccessionVersionId().getVersion())));
+        if (sharedStudy != null) {
+            return sharedStudy;
+        }
+        return persistStudy(study);
```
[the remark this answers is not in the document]
tskir: Sorry, I'm not sure if I understood your concern correctly. Currently there are two DB queries: first attempts to find an existing Study, second persists it if necessary. Without the first query we can't be sure that the same Study hasn't already been imported which will lead to duplicates. Besides, the code for `MetadataStudyFinderOrPersister.java` is @selvaebi's, not mine :)

tskir:
Rebased on top of recent changes in #47. Addressed review comments. **Ready for the fifth iteration of review.** [*Diff link*](https://github.com/selvaebi/metadata-ws/compare/T2D-287StoreAnalysis...ebivariation:t2d-297-retrieve-reference-sequence)

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -142,6 +162,55 @@ public Analysis importAnalysis(String accession) {
...
+    }
+
+    private Set<String> getReferenceSequenceAccessions(AnalysisType analysisType) {
+        Set<String> referenceSequenceAccessions = new HashSet<>();
+        AnalysisType.ANALYSISTYPE analysisType1 = analysisType.getANALYSISTYPE();
+        // ENA records can contain a reference sequence in either of the three analysis categories:
+        // REFERENCE_ALIGNMENT, SEQUENCE_VARIATION, and PROCESSED_READS. All accessions are collected and returned.
+        if (analysisType1.isSetREFERENCEALIGNMENT()) {
+            referenceSequenceAccessions.add(
+                    // TODO: should we use getRefname() or getAccession()?
```
selvaebi: All EGA xmls have refName only

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/configuration/MetadataImporterMainApplicationConfiguration.java
```diff
@@ -47,34 +51,59 @@
...
-            FileRepository fileRepository,
-            TaxonomyRepository taxonomyRepository,
-            AnalysisRepository analysisRepository,
-            StudyRepository studyRepository) {
-        return new SraObjectsImporterThroughDatabase(sraXmlRetrieverThroughDatabase, sraStudyXmlParser(),
-                studyConverter(), publicationExtractorFromStudy(publicationRepository),
-                webResourceExtractorFromStudy(webResourceRepository), taxonomyExtractor(taxonomyRepository),
-                sraAnalysisXmlParser(), analysisConverter(), fileExtractorFromAnalysis(fileRepository),
-                analysisRepository,studyRepository);
+    public ObjectsImporter objectImporterThroughEnaDatabase(SraXmlRetrieverThroughDatabase sraXmlRetrieverThroughDatabase,
```
selvaebi: SraXmlRetrieverThroughDatabase can be moved to the next line.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/repositories/ReferenceSequenceCustomRepository.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.List;
+
+@NoRepositoryBean
+public interface ReferenceSequenceCustomRepository extends PagingAndSortingRepository<ReferenceSequence, Long>,
+        QueryDslPredicateExecutor<ReferenceSequence> {
+
+    ReferenceSequence findByAccessions(@Param("accessions") List<String> accessions);
+
+    default ReferenceSequence findOrSave(ReferenceSequence referenceSequence) {
+        /* The below find query will make sure to return the same reference sequence when analyses sharing same
```
selvaebi: i guess we can remove the comment as metadata-ws doesn't need to know about import here.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/repositories/ReferenceSequenceCustomRepository.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.data.querydsl.QueryDslPredicateExecutor;
+import org.springframework.data.repository.NoRepositoryBean;
+import org.springframework.data.repository.PagingAndSortingRepository;
+import org.springframework.data.repository.query.Param;
+import uk.ac.ebi.ampt2d.metadata.persistence.entities.ReferenceSequence;
+
+import java.util.List;
+
+@NoRepositoryBean
+public interface ReferenceSequenceCustomRepository extends PagingAndSortingRepository<ReferenceSequence, Long>,
```
selvaebi:
```suggestion
public interface ReferenceSequenceRepositoryCustom extends PagingAndSortingRepository<ReferenceSequence, Long>,
```

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
selvaebi: so I guess we can just use the refName to create the referenceSequence in SraObjectImportThroughDatabase

tskir: @selvaebi Thank you for the clarification. I changed the implementation so that `getAccession()` is used by default while `getRefname()` is used for import from the database.

@cyenyxe What should we do in the case of ENA? Consider the two following examples:
* Analysis [ERZ496533](https://www.ebi.ac.uk/ena/data/view/ERZ496533&display=xml), which is a `SEQUENCE_VARIATION`, has assembly [GCA_000002305.1](https://www.ebi.ac.uk/ena/data/view/GCA_000002305.1&display=xml) which only has an accession. This is a standard assembly which can be parsed & used as usual.
* Analysis [ERZ004000](https://www.ebi.ac.uk/ena/data/view/ERZ004000&display=xml), which is a `REFERENCE_ALIGNMENT`, has assembly [HAAA01000000](https://www.ebi.ac.uk/ena/data/view/HAAA01000000&display=xml), which only has a refName. Moreover, this is not a standard assembly; it returns an <ENTRY> entity, and I'm not sure what to do with them (and whether we _want_ to do anything with them)

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/configuration/MetadataImporterMainApplicationConfiguration.java
tskir: Fixed, thank you

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/repositories/ReferenceSequenceCustomRepository.java
tskir: Makes sense, done

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/repositories/ReferenceSequenceCustomRepository.java
tskir: Renamed, thanks

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/MetadataImporterMainApplication.java
```diff
@@ -45,22 +44,23 @@
...
-
-    public MetadataImporterMainApplication(ObjectsImporter objectsImporter, StudyRepository studyRepository) {
+    public MetadataImporterMainApplication(ObjectsImporter objectsImporter) {
         this.objectsImporter = objectsImporter;
-        this.studyRepository = studyRepository;
     }
 
     public static void main(String[] args) throws Exception {
         SpringApplication.run(MetadataImporterMainApplication.class, args);
+
```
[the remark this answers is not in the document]
tskir: Done

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -142,6 +162,55 @@ public Analysis importAnalysis(String accession) {
...
+                    getAccessionFromReferenceSequenceType(analysisType1.getREFERENCEALIGNMENT()));
+        }
+        else if (analysisType1.isSetSEQUENCEVARIATION()) {
```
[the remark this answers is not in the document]
tskir: Noted & fixed, thank you

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -142,6 +162,55 @@ public Analysis importAnalysis(String accession) {
...
+        }
+        else if (analysisType1.isSetPROCESSEDREADS()) {
+            referenceSequenceAccessions.add(
+                    getAccessionFromReferenceSequenceType(analysisType1.getPROCESSEDREADS()));
+        }
+        referenceSequenceAccessions.remove(null);
+        return referenceSequenceAccessions;
+    }
+
+    private String getAccessionFromReferenceSequenceType(ReferenceSequenceType referenceSequenceType){
```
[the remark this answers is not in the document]
tskir: Fixed

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -142,6 +162,55 @@ public Analysis importAnalysis(String accession) {
...
+        }
+        else if (analysisType1.isSetSEQUENCEVARIATION()) {
+            referenceSequenceAccessions.add(
+                    getAccessionFromReferenceSequenceType(analysisType1.getSEQUENCEVARIATION()));
+        }
+        else if (analysisType1.isSetPROCESSEDREADS()) {
+            referenceSequenceAccessions.add(
+                    getAccessionFromReferenceSequenceType(analysisType1.getPROCESSEDREADS()));
+        }
+        referenceSequenceAccessions.remove(null);
```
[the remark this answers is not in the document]
tskir: Agreed; and I also noticed an implementation flaw in that is should have been not `if / else if / else if` block, but rather just a series of `if`s. I'm not sure whether an ENA analysis can include multiple types in practice, but the parser class implementation (e. g. method is named `isSetREFERENCEALIGNMENT`, not `isREFERENCEALIGNMENT`) suggests that theoretically it can happen.

Rewrote this block to be a series of `if`s and also to avoid removing `null` afterwards

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/configuration/MetadataImporterMainApplicationConfiguration.java
```diff
@@ -47,34 +51,59 @@
...
-                                                       FileRepository fileRepository,
-                                                       TaxonomyRepository taxonomyRepository,
-                                                       AnalysisRepository analysisRepository,
-                                                       StudyRepository studyRepository) {
-        return new SraObjectsImporterThroughAPI(sraXmlRetrieverThroughApi, sraStudyXmlParser(),
-                studyConverter(), publicationExtractorFromStudy(publicationRepository),
-                webResourceExtractorFromStudy(webResourceRepository), taxonomyExtractor(taxonomyRepository),
-                sraAnalysisXmlParser(), analysisConverter(), fileExtractorFromAnalysis(fileRepository),
-                analysisRepository, studyRepository);
+                                                   PublicationRepository publicationRepository,
```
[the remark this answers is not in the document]
tskir: Hmm, it's really strange I didn't notice this. Now fixed!

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/configuration/MetadataImporterMainApplicationConfiguration.java
```diff
@@ -47,34 +51,59 @@
...
+                fileExtractorFromAnalysis(fileRepository),
+                sraAssemblyXmlParser(),
+                referenceSequenceConverter(),
+                analysisRepository,
+                studyRepository,
+                referenceSequenceRepository
+        );
     }
 
     private StudyConverter studyConverter() {
```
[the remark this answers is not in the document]
tskir: I noticed this problem as well and have already performed a conceptual grouping in #55. I suggest that in this PR we leave this as is to avoid unnecessary downstream rebasing.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/converter/ReferenceSequenceConverter.java
```diff
@@ -0,0 +1,38 @@
+/*
+ *
+ * Copyright 2018 EMBL - European Bioinformatics Institute
```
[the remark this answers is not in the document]
tskir: Fixed. Maybe we should do a global replace in all files to make it "2018-2019"?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/converter/ReferenceSequenceConverter.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class ReferenceSequenceConverter implements Converter<AssemblyType, ReferenceSequence> {
+
+    @Override
+    public ReferenceSequence convert(AssemblyType assemblyType) {
+        return new ReferenceSequence(
+                assemblyType.getNAME(),
+                "NOT_SPECIFIED",  // ENA only specifies accession+version, not the patch
+                Arrays.asList(assemblyType.getAccession()),
+                ReferenceSequence.Type.ASSEMBLY  // ENA data model only has ASSEMBLY type
```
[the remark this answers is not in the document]
tskir: I've looked through the entire `uk.ac.ebi.ena.sra.xml` package but could't find anything resembling these types. I also compiled a [spreadsheet](https://docs.google.com/spreadsheets/d/18IT7defkFXC0zkwhO0kQyGWOi1UbMmN9SjJq6B3yHfc/edit?usp=sharing) of all possible analysis types with accession examples, and in my experience they either contain a `ReferenceSequenceType`, which in turn links to an `ASSEMBLYType` (and _only_ to it), or they contain nothing at all. If you could provide me with a pointer, e. g. an accession example where you think the reference sequence is not an assembly, but instead a gene sequence or a transcriptome, I would be thankful.

This may be related to the refName/accession thing (see discussion [here](https://github.com/EBIvariation/metadata-ws/pull/50#discussion_r287836865)), because one of the analysis accessions returns something weird which is not an Assembly (https://www.ebi.ac.uk/ena/data/view/HAAA01000000&display=xml). May this be what we're looking for?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/TaxonomyExtractorFromReferenceSequence.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ */
+
+package uk.ac.ebi.ampt2d.metadata.importer.extractor;
+
+import uk.ac.ebi.ampt2d.metadata.persistence.entities.Taxonomy;
+import uk.ac.ebi.ampt2d.metadata.persistence.repositories.TaxonomyRepository;
+import uk.ac.ebi.ena.sra.xml.AssemblyType;
+
+public class TaxonomyExtractorFromReferenceSequence {
```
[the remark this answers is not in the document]
tskir: I can't _yet_ rename it to `TaxonomyExtractor` because this name is taken by a dummy taxonomy generator, which will only be removed when we merge #54.

As for the class structure:
* The taxonomy extraction & persistence was originally implemented as a separate class;
* Then, following review comments and changing upstream class structure, moved inside the ReferenceSequence persistence logic;
* Then moved to a separate class _again_ as per @selvaebi's suggestion.

Currently it is modelled after `PublicationExtractorFromStudy` and similar other classes. Could you please elaborate on how I should change it and where to move its code?

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/importer/converter/ReferenceSequenceConverterTest.java
```diff
@@ -0,0 +1,82 @@
...
+        xmlParser = new SraAssemblyXmlParser();
+        referenceSequenceConverter = new ReferenceSequenceConverter();
+    }
+
+    @Test
+    public void convertFromApiXml() throws Exception {
+        String assemblyAccession = "GCA_000002305.1";
+        AssemblyType assemblyType = getAssemblyType(ASSEMBLY_DOCUMENT_API_XML, assemblyAccession);
+        ReferenceSequence referenceSequence = referenceSequenceConverter.convert(assemblyType);
+        Assert.assertNotNull(referenceSequence);
```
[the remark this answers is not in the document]
tskir: Fixed

tskir:
Addressed all review comments. **Ready for the sixth iteration of review.** [Diff link](https://github.com/EBIvariation/metadata-ws/pull/50/files)

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
cyenyxe: We should only process `SEQUENCE_VARIATION` analyses for now, but your question about HAAA01000000 still applies.

That is a transcriptome, one of the 3 types of sequences we support, so it has to be processed. The set of sequences it contains is
> \<xref db="ENA-TSA" id="HAAA01000001-HAAA01003524"/\>

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/converter/ReferenceSequenceConverter.java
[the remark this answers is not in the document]
cyenyxe: Not really, given that we copyright them individually, each of them should start in the creation date.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/TaxonomyExtractorFromReferenceSequence.java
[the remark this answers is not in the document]
cyenyxe: I guess my concern is due to the fact that this class doesn't really deal with a reference sequence at any point, just with taxonomies, and that is why I see it been closer to `StudyRepository`. If that was the case, we would need to replicate the `StudyRepositoryCustom(Impl)` class hierarchy, the `findOrCreate` method in the implementation instead of the top interface declaration.

cyenyxe:
Have the changes been pushed? I still see the old version of the code.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -142,6 +162,55 @@ public Analysis importAnalysis(String accession) {
...
+        return referenceSequenceAccessions;
+    }
+
+    private String getAccessionFromReferenceSequenceType(ReferenceSequenceType referenceSequenceType){
+        if (referenceSequenceType == null) {
```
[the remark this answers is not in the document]
tskir: Thanks, fixed

tskir:
> Have the changes been pushed? I still see the old version of the code.

Yep, I indeed forgot to push it >< Now fixed as well as the latest suggestion from @iyangar. Ready for review

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/MetadataImporterMainApplication.java
```diff
@@ -52,7 +52,6 @@ public MetadataImporterMainApplication(ObjectsImporter objectsImporter) {
     public static void main(String[] args) throws Exception {
         SpringApplication.run(MetadataImporterMainApplication.class, args);
     }
```
cyenyxe:
```suggestion
    }

```

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -142,9 +162,64 @@ public Analysis importAnalysis(String accession) {
...
+            IMPORT_LOGGER.log(Level.SEVERE, exception.getMessage());
+        }
+        return referenceSequence;
+    }
+
+    private Set<String> getReferenceSequenceAccessions(AnalysisType analysisType) {
+        Set<String> referenceSequenceAccessions = new HashSet<>();
+        AnalysisType.ANALYSISTYPE analysisType1 = analysisType.getANALYSISTYPE();
+        // ENA records can contain a reference sequence in either of the three analysis categories:
+        // REFERENCE_ALIGNMENT, SEQUENCE_VARIATION, and PROCESSED_READS. All accessions are collected and returned.
```
cyenyxe: The EVA and AMP only deal with SEQUENCE_VARIATION analyses, could @jdylan please confirm if EGA needs others?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
```diff
@@ -118,6 +119,11 @@ protected Study extractAnalysisFromStudy(StudyType studyType, Study study) {
         return study;
     }
 
+    @Override
+    protected String getAccessionFromStandard(ReferenceAssemblyType.STANDARD standard) {
```
cyenyxe: It is counter-intuitive that a method expected to return the accession in fact doesn't. What is the reason to return always the refname when importing from the database?

selvaebi: it should be for EGA case, maybe if the object has accession call super. import else, need to create RefSeq from refname.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/converter/ReferenceSequenceConverter.java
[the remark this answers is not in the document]
tskir: Then, presumably, we also need to update the years for modules which were created in 2018 and then edited in 2019, i. e. “Copyright 2018” → “Copyright 2018–2019”? I suspect we'll generally forget to maintain this :)

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/MetadataImporterMainApplication.java
tskir: Thanks, fixed

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
tskir: By the way, should metadata API support ENA as well eventually, or it's not planned, ever?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
tskir: > It is counter-intuitive that a method expected to return the accession in fact doesn't

It is, but the reason we have to do this, again if I understand correctly, because sometimes refName is used as a _de facto_ accession in EGA, and we have to use it to identify the reference sequence. Maybe @selvaebi and @jdylan could either confirm or correct me.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
cyenyxe: The API doesn't need to support ENA.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
cyenyxe: Even if EGA uses the reference name right now, that can be easily mapped to a proper accession, for instance:

* GRCh37 would have the accession GCA_000001405.1
* GRCh38 would have the accession GCA_000001405.15
* GRCh38.p13 (latest) would have the accession GCA_000001405.28

We should always favor accessions because they are guaranteed to be unique.

iyangar: Are we confined to only above three vi.z., GRCh37, GRCh38 and GRCh38.p13. 
Should we consider other reference name like `GRCh37.p13` with accession `GCA_000001405.14`

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
iyangar: Will remove `REFERENCE_ALIGNMENT` and `PROCESSED_READS`

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -142,9 +162,64 @@ public Analysis importAnalysis(String accession) {
...
+                if (standard != null) {
+                    return getAccessionFromStandard(standard);
+                }
+            }
+        }
         return null;
     }
 
+    protected String getAccessionFromStandard(ReferenceAssemblyType.STANDARD standard) {
+        return standard.getAccession();
```
iyangar: @cyenyxe just wanna confirm this is fine? 
[ObjectImporter::getAccessionFromStandard() and SraObjectsImporterThroughDatabase::getAccessionFromStandard()   https://github.com/EBIvariation/metadata-ws/pull/50#discussion_r288505075 ]

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
cyenyxe: Yes, we need to consider from GCA_000001405.1 to GCA_000001405.28.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
cyenyxe: Has @jdylan confirmed that the EGA is not using them?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
cyenyxe: Yes, it must always return the accession.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
iyangar: I interpreted https://github.com/EBIvariation/metadata-ws/pull/50#discussion_r288970412 so. Will wait for formal confirmation from @jdylan

selvaebi: I could see from the xmls EGA only uses SEQUENCE_VARIATION. but anyways Dylan can confirm.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
iyangar: Mapping case-sensitive reference name to accession

- GRCh37 through GRCh37.p13 = GCA_000001405.1 through GCA_000001405.14

- GRCh38 through GRCh38.p13 = GCA_000001405.15 through GCA_000001405.28

Ref: https://www.ncbi.nlm.nih.gov/assembly/?term=GRCh37 and https://www.ncbi.nlm.nih.gov/assembly?term=GRCh38&cmd=DetailsSearch

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
```diff
@@ -121,7 +125,131 @@ protected Study extractAnalysisFromStudy(StudyType studyType, Study study) {
 
     @Override
     protected String getAccessionFromStandard(ReferenceAssemblyType.STANDARD standard) {
-        return standard.getRefname();
+        String refName = standard.getRefname();
+        String accession = null;
+
+        try {
+            switch (refName) {
```
cyenyxe: Instead of making this effectively a linear search through the use of a switch, you can create a static (initialized just once) Map and store the mappings there. That will make every search have O(1) complexity.

I would also suggest making the search case insensitive, we have seen `grch37` or similar used many times.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
```diff
@@ -126,130 +157,7 @@ protected Study extractAnalysisFromStudy(StudyType studyType, Study study) {
...
-                case "GRCh38.p12":
-                    accession = "GCA_000001405.27";
-                    break;
-
-                case "GRCh38.p13":
-                    accession = "GCA_000001405.28";
-                    break;
-
-                default:
-                    throw new Exception("Encountered exception for unknown reference sequence name " + refName);
```
cyenyxe: This exception still applies, using a Map doesn't prevent the user from requesting an unknown reference. But it should be an `IllegalArgumentException` instead of a completely generic `Exception`.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
```diff
@@ -126,130 +157,7 @@ protected Study extractAnalysisFromStudy(StudyType studyType, Study study) {
...
-            }
-        } catch (Exception e) {
-            IMPORT_LOGGER.log(Level.SEVERE, e.getMessage());
-        }
-
-        return accession;
+        return REFERENCE_MAP.get(refName.toLowerCase());
```
selvaebi: please add a check for refName !=null

selvaebi:
please add test cases as well.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -159,6 +199,41 @@ public Analysis importAnalysis(String accession) {
...
+            ReferenceSequenceType referenceSequenceType = analysistype.getSEQUENCEVARIATION();
+            if (referenceSequenceType != null) {
+                ReferenceAssemblyType referenceAssemblyType = referenceSequenceType.getASSEMBLY();
+                if (referenceAssemblyType != null) {
+                    ReferenceAssemblyType.STANDARD standard = referenceAssemblyType.getSTANDARD();
+                    if (standard != null) {
+                        String refName = standard.getRefname();
+                        String accession = null;
+                        if (refName != null) {
+                            accession = REFERENCE_MAP.get(refName.toLowerCase());
```
selvaebi: this manual creation is just for database case / ega, so it should be moved to SraDatabaseImporter class

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
iyangar: Need to handle all 3. https://www.ebi.ac.uk/panda/jira/browse/T2D-297?focusedCommentId=281856&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-281856

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -158,20 +238,38 @@ public Sample importSample(String accession) {
...
+            referenceSequenceType = analysistype.getPROCESSEDREADS();
         }
-        ReferenceAssemblyType.STANDARD standard = referenceAssemblyType.getSTANDARD();
-        if (standard != null) {
-            referenceSequenceAccessions.add(standard.getAccession());
+
+        if (referenceSequenceType != null) {
+            ReferenceAssemblyType referenceAssemblyType = referenceSequenceType.getASSEMBLY();
+            if (referenceAssemblyType != null) {
+                ReferenceAssemblyType.STANDARD standard = referenceAssemblyType.getSTANDARD();
```
selvaebi: This all logic of manual creation of ReferenceSequnce should go to SraImporterDatabase

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -118,12 +140,15 @@ public Analysis importAnalysis(String accession) {
             AnalysisType analysisType = sraAnalysisXmlParser.parseXml(xml, accession);
             analysis = analysisConverter.convert(analysisType);
             analysis.setFiles(fileExtractorFromAnalysis.getFiles(analysisType));
+
             List<ReferenceSequence> referenceSequences = new ArrayList<>();
-            for (String referenceSequenceAccession : getReferenceSequenceAccessions(analysisType)) {
-                //TODO ReferenceSequence Import
-                //referenceSequences.add(importReferenceSequence(referenceSequenceAccession));
+            ReferenceSequence referenceSequence = buildReferenceSequence(analysisType);
```
selvaebi: ImportReferenceSequence is not called at all  ?

iyangar: In ImportReferenceSequence() => IsraXmlRetrieverByAccession.getXml is selecting analysis record with reference sequence accession and fails. If i am correct on slack we discussed this and I interpreted DB call should not be for ref seq. Please let me know

selvaebi: yes, you need to override importReferenceSequence in SraImporterDatabase to manually create refSequence instead calling super.

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/importer/api/SraObjectsImporterThroughAPITest.java
```diff
@@ -94,4 +99,16 @@ public void importAnalysisObject() throws Exception {
         assertEquals(0, studyRepository.count());
         assertEquals(0, analysisRepository.count());
     }
-}
\ No newline at end of file
+
+    @Test
+    public void importReferenceSequenceObject() throws Exception {
```
selvaebi: also, add asserts in importStudy to check whether reference sequences are imported?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
iyangar: sure

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
tskir: @iyangar Transcriptome and gene sequences are still not supported in the latest version of the PR. See example of a transcriptome reference sequence here:
* https://www.ebi.ac.uk/ena/data/view/ERZ004000&display=xml
* https://www.ebi.ac.uk/ena/data/view/HAAA01000000&display=xml

For it, we have an empty `STANDARD` (no accession, only a refname), and the actual accessions are directly inside the `REFERENCE_ALIGNMENT`:

```
<SEQUENCE accession="HAAA01000001.1" label="PNEJI2.1.1"/>
<SEQUENCE accession="HAAA01000002.1" label="PNEJI2.2.1"/>
<SEQUENCE accession="HAAA01000003.1" label="PNEJI2.3.1"/>
...
```

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/TaxonomyExtractorFromReferenceSequence.java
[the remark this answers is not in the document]
tskir: @cyenyxe I see your point now. I suppose the best way would be to implement a `TaxonomyRepositoryCustom` with  a `findOrCreate` method and to simply call that at the appropriate point during the reference sequence import

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
tskir: In the latest version of the PR it's implemented slightly different from my original implementation. Originally it combined accessions from all types of analyses if it could find them; now it checks sequentially in a `else-if` chain. If we are guaranteed for sure that two different tags cannot be present in the same analysis (e. g. a `SEQUENCE_VARIATION` and a `REFERENCE_ALIGNMENT`), then it's fine. @cyenyxe @jdylan can you confirm that each analysis contains at most one analysis type?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
iyangar: correct, In today's standup we decided to discuss about this in sprint meeting

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
iyangar: Yes, in today's i confirmed with @jdylan. Analysis can be of only one type SEQUENCE_VARIATION or REFERENCE_ALIGNMENT or PROCESSED_READS

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
iyangar: Is there a way to get reference sequence name from its accession? (From analysisType ref seq accession is retrieved and getting ref name is inside this is straight forward)

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/importer/api/SraObjectsImporterThroughAPITest.java
selvaebi: this is not done yet

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
tskir: In the current implementation all three analysis types are supported in a reasonable manner. Marking as resolved.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
tskir: In the current implementation accession is always returned. Marking as resolved.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
tskir: Fixed by @iyangar; in the current implementation `IllegalArgumentException` is thrown

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
tskir: Fixed by @iyangar, check is now present

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
tskir: This has been fixed & implemented by @iyangar. Now certain human build refnames are converted to accessions

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
tskir: Both points fixed by @iyangar

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
tskir: Fixed by @iyangar, now moved to `SraObjectsImporterThroughDatabase`

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
tskir: This comment was probably made before @iyangar made his latest changes (I'm not sure). Anyway in the latest implementation `getAccessionFromReferenceSequenceType` method should remain in the `ObjectsImporter` class because it is used in both API and database cases. It is used to extract accession from ENA's `ReferenceSequenceType`.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
tskir: @iyangar it should be possible, although accessions are not always present and they are not always RefSeq accessions. Missing refname is one of the edge cases which I outlined in the [spreadsheet](https://docs.google.com/spreadsheets/d/18IT7defkFXC0zkwhO0kQyGWOi1UbMmN9SjJq6B3yHfc/edit?usp=sharing), let's see what decision is made concerning it.

metadata-load/src/test/java/uk/ac/ebi/ampt2d/metadata/importer/api/SraObjectsImporterThroughAPITest.java
tskir: Actually, currently no reference sequences are imported in `importStudy` , because two studies don't have any analyses (SRP000392, SRP000118), and the remaining study ERP000054 has only one analysis ERZ000001 which has a custom (currently non-importable) reference sequence:
```
               <ASSEMBLY>
                    <CUSTOM>
                         <URL_LINK>
                              <LABEL>Mus_musculus.NCBIM37.52</LABEL>
                              <URL>ftp://ftp.ensembl.org/pub/release-52/fasta/mus_musculus/dna/Mus_musculus.NCBIM37.52.dna.toplevel.fa.gz</URL>
                         </URL_LINK>
                    </CUSTOM>
               </ASSEMBLY>
```

To add this test, I replaced `ERP000054` with `ERP006576`, which has two analyses and one reference sequence.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -22,7 +22,6 @@
 import uk.ac.ebi.ampt2d.metadata.importer.extractor.FileExtractorFromAnalysis;
```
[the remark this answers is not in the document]
tskir: This is a valid concern. I suggest we move it to #55, because I have already performed a significant reordering/grouping of methods and class attributes in that PR. If I change anything here, an unnecessary chain of downstream rebases would be required.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
iyangar: This is implemented in SraObjectsImporterThroughDatabase. This comment was made before i implemented by saving both accession and ref name.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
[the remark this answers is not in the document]
cyenyxe: We don't order methods as in public, protected then private. We follow the recommendations from the Clean Code book, where a private method follows the first public method that uses it, in order to keep a particular flow as close as possible.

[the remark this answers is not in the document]
tskir: Fixed in #55, see [discussion](https://github.com/EBIvariation/metadata-ws/pull/55#pullrequestreview-247590754)

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
```diff
@@ -76,7 +137,20 @@ public Analysis importAnalysis(String accession) {
 
     @Override
     public ReferenceSequence importReferenceSequence(String accession) {
-        return super.importReferenceSequence(accession);
+        String[] accnRefNameSplit = accession.split("#", 2);
```
cyenyxe: The split by '#' is not very intuitive (although documented in another method). Could you please add some method documentation to show what the input of this method looks like, and where does it come from?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
```diff
@@ -76,7 +137,20 @@ public Analysis importAnalysis(String accession) {
...
+        String refAccession = accnRefNameSplit[0];
+        String refName = accnRefNameSplit[1];
+        ArrayList<String> accessionList = new ArrayList<>(Arrays.asList(refAccession));
+        String[] refNameSplit = refName.split("\\.", 2);
+        String patch = null;
+        if (refNameSplit.length == 2) {
+            patch = refNameSplit[1];
+        }
+        ReferenceSequence referenceSequence = new ReferenceSequence(refName, patch, accessionList,  ReferenceSequence.Type.ASSEMBLY);
```
cyenyxe:
```suggestion
        ReferenceSequence referenceSequence = new ReferenceSequence(refName, patch, accessionList, ReferenceSequence.Type.ASSEMBLY);
```

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
tskir: Added the documentation in https://github.com/EBIvariation/metadata-ws/pull/50/commits/f992ccc9192cdef417bcfafd9a2ea1b554e32493. This is @iyangar's implementation for refname/accession handling in EGA import, and I think it's pretty reasonable

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/database/SraObjectsImporterThroughDatabase.java
tskir: Fixed in https://github.com/EBIvariation/metadata-ws/pull/50/commits/f992ccc9192cdef417bcfafd9a2ea1b554e32493.

tskir:
> According to GitHub it looks like it wasn't fixed, could you please edit it manually?

Whoops, turned out I fixed this, but in the wrong PR (one of the downstream ones). Now fixed here as well, commit https://github.com/EBIvariation/metadata-ws/pull/50/commits/f992ccc9192cdef417bcfafd9a2ea1b554e32493.

PR #51, opened 2019-04-15
T2D-260 Introduce M:1 relation from ReferenceSequence to Taxonomy

iyangar:


metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/configuration/RepositoryConfiguration.java
```diff
@@ -35,4 +36,9 @@ AnalysisEventHandler analysisEventHandler() {
         return new AnalysisEventHandler();
     }
 
+    @Bean
+    ReferenceSequenceEventHandler referenceSequenceEventHandler() {
```
Zhicheng-Liu: Do we really need the event handler to validate the taxonomy? Is `@ManyToOne(optional=false)` not enough?

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/repositories/ReferenceSequenceRepository.java
```diff
@@ -36,4 +36,13 @@ default void customize(QuerydslBindings bindings, QReferenceSequence referenceSe
                 .first((path, value) -> path.any().equalsIgnoreCase(value.iterator().next()));
     }
 
+
+    /*
```
Zhicheng-Liu: Please remove these lines.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/ReferenceSequenceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.List;
+
+public class ReferenceSequenceImpl implements ReferenceSequenceService {
+
+    @Autowired
+    private ReferenceSequenceRepository referenceSequenceRepository;
+
+    @Override
+    public List<ReferenceSequence> findReferenceSequencesByPredicate(Predicate predicate) {
```
Zhicheng-Liu: It looks like this method is not currently used by any client code. Can we mark it `private` and remove it from the interface?

metadata-ws/src/test/resources/features/study.feature
```diff
@@ -72,14 +72,14 @@ Feature: study object
 
     Examples:
       | base | query | N | url |
-      | taxonomy-id | id=9606 | 1 | STUDY1 |
+#      | taxonomy-id | id=9606 | 1 | STUDY1 |
```
Zhicheng-Liu: Why these "examples" are commented out?

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/configuration/RepositoryConfiguration.java
iyangar: When invalid links are sent and/or patch happens this will not catch. Hence even handler is reqiured

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/ReferenceSequenceImpl.java
iyangar: Implementing interface requires the method to be public

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/configuration/RepositoryConfiguration.java
Zhicheng-Liu: When invalid links are used (as in "Scenario: update a reference sequence with invalid taxonomy and fail"), the exception is caught, but not handled: https://github.com/EBIvariation/metadata-ws/blob/master/metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/exceptionhandling/ExceptionHandlers.java#L43

My point was if there is nothing this event handler does that other exception handler can't do then we should avoid duplication. We can't add event handler for each madantory link.

iyangar: It can be done. But we need to pass generic message to user.
Otherwise we need to parse strings, where the format might change with spring versions

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/events/ReferenceSequenceEventHandler.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @HandleBeforeLinkDelete
+    @HandleBeforeLinkSave
+    public void validateTaxonomies(ReferenceSequence referenceSequence, Object taxonomy) {
+        validateTaxonomyLink(referenceSequence);
+    }
+
+    private void validateTaxonomyLink(ReferenceSequence referenceSequence) {
+        if (referenceSequence.getTaxonomy() == null) {
```
selvaebi: This can be checked using @NotNull in the pojo

iyangar: Yes, I and Zhi had discussion on this https://github.com/EBIvariation/metadata-ws/pull/51#discussion_r275762826. 
Should we pass specific message or generic message to user.

selvaebi: @NotNull will give u a default exception message that is far more enough I guess.

iyangar: There is are 2 choices
1. @Notnull
This works best when there is no link for taxonomy.
When invalid Taxonomy link is passed, handler will strip-off invalid link and @NotNull will throw error "Taxonomy field can be null".
2. Handle in ExceptionHandlers.java 
Error code can be verified here; pass the message `Null or invalid value is provided` (without Taxonomy word) to user. Since it applies to all such constraints it can not be confined to Taxonomy

tskir: I, too, feel that adding an event handler for just a linking constraint violation is excessive. Especially given that the error irrecoverable anyway, i. e. there is no scenario where we would want to catch `ReferenceSequenceWithoutValidTaxonomyException` and do something other than informing the user. `@NotNull(message="A reference sequence must have one valid taxonomy")` will do just fine.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/ReferenceSequenceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.querydsl.core.types.Predicate;
+import org.springframework.beans.factory.annotation.Autowired;
+import uk.ac.ebi.ampt2d.metadata.persistence.entities.QReferenceSequence;
+import uk.ac.ebi.ampt2d.metadata.persistence.entities.ReferenceSequence;
+import uk.ac.ebi.ampt2d.metadata.persistence.repositories.ReferenceSequenceRepository;
+
+import java.util.List;
+
+public class ReferenceSequenceImpl implements ReferenceSequenceService {
```
Zhicheng-Liu: `ReferenceSequenceServiceImpl` please!

PR #52, opened 2019-05-04
Consistent URLs to link between objects

iyangar:
Adding autogenerated id to publication

### DIFF
--- a/metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Publication.java
+++ b/metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Publication.java
@@ -30,7 +30,7 @@ import javax.validation.constraints.NotNull;
 @Entity
 public class Publication extends Auditable<Long>{
 
-    @ApiModelProperty(position = 1, value = "Publication auto generated id", required = true, readOnly = true)
+    @ApiModelProperty(position = 1, value = "Publication auto generated id", readOnly = true)
     @JsonProperty(access = JsonProperty.Access.READ_ONLY)
     @Id
     @GeneratedValue(strategy = GenerationType.AUTO)
--- a/metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Publication.java
+++ b/metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Publication.java
@@ -36,11 +36,18 @@ public class Publication extends Auditable<Long>{
     @GeneratedValue(strategy = GenerationType.AUTO)
     private long id;
 
-    @ApiModelProperty(position = 2, example = "PubId123")
+    @ApiModelProperty(position = 2, example = "PubmedId")
     @NotNull
     @JsonProperty
     @Column(unique = true)
     private String publicationId;
+
+    public Publication() {
+    }
+
+    public Publication(String publicationId) {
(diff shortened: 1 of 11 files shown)


metadata-ws/src/test/resources/features/publication.feature
```diff
@@ -0,0 +1,32 @@
...
+    """
+    Then set the URL to PUBLICATION
+    And the response code should be 201
+    And the Location header should be present with value of PUBLICATION
+    When I request GET with value of PUBLICATION
+    Then the response code should be 200
+    And the response JSON should be:
+    """
+    {
+      "publicationId": "Publication1"
```
selvaebi: why is id and lastModifieddate not visible in json ?

iyangar: Its part of object. Both are auto generated and we dont explicitly include (same for other objects).
Let me know, i can include to verify id as number and lastModifieddate exists

selvaebi: if its a part of the object it should be visible. and also the ticket is adding auto-id, so it should be visible.

selvaebi:
please rebase

metadata-ws/src/test/resources/features/publication.feature
```diff
@@ -0,0 +1,30 @@
+Feature: publication object
+
+  Scenario: register a publication successfully
+    When I request POST /publications with JSON payload:
+    """
+    {
+      "publicationId": "Publication1"
+    }
+    """
+    Then set the URL to PUBLICATION
```
cyenyxe: This is still part of what you do, not the expected outcome, so it should be `and` instead of `then`.

iyangar: Corresponding to this fix, i have done minor changes

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Publication.java
```diff
@@ -20,28 +20,37 @@
...
 import javax.validation.constraints.NotNull;
 
 @Entity
-public class Publication extends Auditable<String>{
+public class Publication extends Auditable<Long>{
 
-    @ApiModelProperty(position = 1, value = "PubmedId", required = true)
-    @JsonProperty
-    @NotNull
+    @ApiModelProperty(position = 1, value = "Publication auto generated id", required = true, readOnly = true)
```
selvaebi: required=true should be removed

iyangar: Sure, i observe required=true is present across all entities. So are we planning to avoid this in future for all entities?

selvaebi: yes we kinda need to remove it for all entitites

selvaebi: can you please remove for all entities for consistency

cyenyxe:
Given that a new publications feature file has been created in the PR, shouldn't the Java code implementing the scenario be included too?

iyangar:
https://github.com/EBIvariation/metadata-ws/pull/52#issuecomment-497710752: All of the java code implementation is part of CommonSteps.java present in repo. Also there is no change required to it.

PR #60, opened 2019-06-21
T2D-312 removed Webresource Type

selvaebi:


metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/WebResource.java
```diff
@@ -38,52 +36,23 @@
...
-
-        CONSORTIUM_WEB,
-
-    }
-
     @ApiModelProperty(position = 1, value = "Web resource auto generated id", readOnly = true)
     @JsonProperty(access = JsonProperty.Access.READ_ONLY)
     @Id
-    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator="WEB_RESOURCE_SEQ")
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "WEB_RESOURCE_SEQ")
```
cyenyxe: We don't really need a sequence generator, we could follow the AUTO strategy.

selvaebi: the AUTO strategy actually chooses Sequence I guess if I don't specify a sequence it uses the same sequence for all entities. let me confirm it again

cyenyxe: Discussed offline that, even though we don't intend these IDs to be externally used, this behavior is not harmful so we will leave it as is.

PR #66, opened 2019-07-30
Fix for secured API access

iyangar:


metadata-ws/src/main/resources/application.properties
```diff
@@ -49,4 +49,5 @@ swagger.apiinfo.termsOfServiceUrl=
 
 security.enabled=true
 security.oauth2.resource.filter-order=3
-security.oauth2.resource.jwt.key-value=@metadata-jwtKey@
+security.metadata.jwtkey=@metadata-jwtKey@
+security.oauth2.resource.jwt.key-value=-----BEGIN PUBLIC KEY-----\n${security.metadata.jwtkey}\n-----END PUBLIC KEY-----
```
cyenyxe: Could the placeholder be written as `@metadata-jwtKey@` in this line too? Why do we need both properties? Why wouldn't the whole string including BEGIN/END PUBLIC KEY work?

iyangar: This workaround is for Jenkins limitation. Jenkins will not pass parameters with space. Hence actual JWT key is taken from jenkins and BEGIN and END are appended to form security.oauth2.resource.jwt.key-value as expected

iyangar: Everything works good

PR #67, opened 2019-07-30
T2D-335 fix jenkins ddlbehaviour

selvaebi:


selvaebi:
the build fails because of a test which imports an analysis where the XML has been edited to include sequence accession. the fix for this is in other PRs.

PR #68, opened 2019-08-09
T2D-338 taxonomy parent

selvaebi:


metadata-ws/src/test/resources/application.properties
```diff
@@ -23,3 +23,5 @@ swagger.apiinfo.termsOfServiceUrl=
 endpoints.entities.release-date.restricted=true
 
 security.enabled=true
+
+spring.datasource.url=jdbc:h2:mem:test;MODE=PostgreSQL
```
iyangar: Is this necessary

selvaebi: yes it is necessary for native query

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/SwaggerConfig.java
```diff
@@ -92,7 +92,8 @@ public Docket metadataApi() {
                         new Tag("Sample Entity", "Sample metadata"),
                         new Tag("Study Entity", "Study metadata"),
                         new Tag("Taxonomy Entity", "Taxonomy definitions"),
-                        new Tag("WebResource Entity", "Study resources that can be accessed through web protocols")
+                        new Tag("WebResource Entity", "Study resources that can be accessed through web protocols"),
+                        new Tag("TaxonomyTree Entity", "Taxonomy Tree")
```
cyenyxe: To follow the capitalization style:
```suggestion
                        new Tag("TaxonomyTree Entity", "Taxonomy tree")
```

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/TaxonomyTree.java
```diff
@@ -0,0 +1,110 @@
...
+public class TaxonomyTree {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "TAXONOMY_TREE_SEQ")
+    private long id;
+
+    @ApiModelProperty(position = 2, dataType = "java.lang.String", notes = "Url to a taxonomySpecies")
+    @JoinColumn(name = "species_id", nullable = false, referencedColumnName = "taxonomyId", unique = true)
+    @OneToOne
+    private Taxonomy taxonomySpecies;
```
cyenyxe: What if we just call the attribute `species`? If we are in a taxonomy tree it should be obvious that the species is associated with a taxonomy, and that would also make unnecessary to set the property `@JoinColumn(name=...)`.

Applies to the rest of columns.

metadata-ws/src/test/resources/features/reference-sequence.feature
```diff
@@ -329,24 +338,34 @@ Feature: reference sequence
 
   Scenario Outline: search various reference sequences by taxonomy name and id
```
cyenyxe: At this point I think it would be more clear to use real cases instead of TAXONOMY_UNRELATEDNUMBERS.

metadata-ws/src/test/resources/features/accession.feature
```diff
@@ -6,7 +6,8 @@ Feature: accession object
     """
     {
       "taxonomyId": 9606,
-      "name": "Homo Sapiens"
+      "name": "Homo Sapiens",
+      "rank":"SPECIES"
```
cyenyxe: To follow spacing style:
```suggestion
      "rank": "SPECIES"
```
There are quite a few occurrences of this.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/TaxonomyTree.java
selvaebi: the problem is with the naming of taxonomy class, we can't name it as class so included taxonomy to all variables.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -305,7 +305,7 @@ public Sample importSample(String accession) {
 
     private Taxonomy extractTaxonomyFromSample(SampleType sampleType) {
         SampleType.SAMPLENAME sampleName = sampleType.getSAMPLENAME();
-        return new Taxonomy(sampleName.getTAXONID(), sampleName.getSCIENTIFICNAME());
+        return new Taxonomy(sampleName.getTAXONID(), sampleName.getSCIENTIFICNAME(), Taxonomy.Rank.SPECIES);
```
tskir: The problem is that we can't be 100% sure that the `getTAXONID()` will return a **species** rank ID. It may also return a subspecies (I think you found an example of that, could you post it here?) or even a "no rank" lower level, especially in case of bacteria and other poorly classified organisms. For example, a submitter might want to specify that the data is from _Staphylococcus aureus 04-02981_ (a specific strain), which is a "no rank" taxon: https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=703339&lvl=3&lin=f&keep=1&srchmode=1&unlock

I suggest that we separate taxonomy information into two parts:
* Original taxonomy ID, stored as a plain integer
* Fixed-rank IDs for species, genus, etc., for indexing and searching

We can discuss this during standup tomorrow.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Taxonomy.java
```diff
@@ -84,8 +91,10 @@ public String getName() {
      * Release date control: taxonomies are always public.
      */
     @Override
+    @JsonIgnore
     public LocalDate getReleaseDate() {
         return null;
     }
 
+    public enum Rank {SUBSPECIES, SPECIES, GENUS, ORDER, CLASS}
```
tskir: Following my comments above, I think that subspecies should be excluded from this list because it is rarely used and is definitely not necessary for indexing

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/TaxonomyTree.java
```diff
@@ -0,0 +1,149 @@
...
+                .collect(Collectors.toList());
+    }
+
+    public List<Taxonomy> getTaxonomySubSpecieses() {
+        return taxonomySubSpecieses;
+    }
+
+    public void setTaxonomySubSpecieses(List<Taxonomy> taxonomySubSpecieses) {
+        this.taxonomySubSpecieses = taxonomySubSpecieses;
+    }
```
tskir: The getters/setters are getting really repetitive, maybe we'll want to use something like this https://projectlombok.org/features/GetterSetter to get rid of this?

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/TaxonomySteps.java
```diff
@@ -59,4 +55,38 @@ public void performPostOnTaxonomies(long id, String name, String ancestorKeys) t
...
+                "\"taxonomyGenus\": \"" + CommonStates.getUrl(genusUrl) + "\"," +
+                "\"taxonomyOrder\": \"" + CommonStates.getUrl(orderUrl) + "\"," +
+                "\"taxonomyClass\": \"" + CommonStates.getUrl(classUrl) + "\"," +
+                "\"taxonomySubSpecieses\":" + objectMapper.writeValueAsString(CommonStates.getUrls(subspeciesUrls))
+                + "}";
+        CommonStates.setResultActions(mockMvc.perform(post("/taxonomyTrees")
+                .with(CommonStates.getRequestPostProcessor())
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(jsonContent.getBytes())));
+    }
```
tskir: Do I understand correctly that if someone wants to create a new taxonomy, they will have to first create all intermediate ranks (species/genus/order/class), and only _then_ create a taxonomy? This is a problem because it makes it possible to accidentally specify an incorrect taxonomy tree. A better approach would be to request an original taxonomy ID (species or a lower level) and then create/substitute all other levels automatically

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
selvaebi: ya I forgot to remove the rank over here, I could just return the taxonomyId for the later use.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/TaxonomyTree.java
selvaebi: Lombak is good to reduce the boilerplate code, maybe we can try in other ticket to check the vulnerabilities. we used Lombok long before and we faced issues in check style source code tools and was dependent on IDEs. but those won't be a problem now I guess.

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/TaxonomySteps.java
selvaebi: this is creating a new taxonomy tree not creating a taxonomy. Taxonomy can be created independently but for taxonomyTree, u need the dependent ranking taxonomies

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/TaxonomySteps.java
```diff
@@ -59,4 +55,38 @@ public void performPostOnTaxonomies(long id, String name, String ancestorKeys) t
...
+                "}";
+        CommonStates.setResultActions(mockMvc.perform(post("/taxonomyTrees")
+                .with(CommonStates.getRequestPostProcessor())
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(jsonContent.getBytes())));
+    }
+
+    @When("^I request POST taxonomyTree with (.*) for species , (.*) for GENUS , (.*) for ORDER , (.*) for CLASS and " +
+            "(.*) for SUBSPECIES$")
+    public void performPostOnTaxonomyTreeWithSubspecies(String speciesUrl, String genusUrl, String orderUrl, String
```
iyangar: are these URLs or urlKeys? If so you name speciesUrlKey
Same others

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
tskir: As far as I can see the taxonomy is still created with a fixed rank of "SPECIES", even though it can be something else

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/StudyServiceImpl.java
```diff
@@ -93,20 +96,22 @@ public int compare(Study o1, Study o2) {
 
     @Override
     public List<Study> findStudiesByTaxonomyId(long id) {
-        QStudy study = QStudy.study;
-        Predicate predicate = study.taxonomy.taxonomyId.eq(id).
-                or(study.taxonomy.ancestors.any().taxonomyId.eq(id));
-
-        return findStudiesByPredicate(predicate);
+        List<Long> speciesAndSubspeciesTaxonomyIds =
+                taxonomyTreeService.findAllSpeciesAndSubspeciesTaxonomyIdsInATaxonomyTreeByTaxonomyId(id);
```
tskir: There are still references to "species and subspecies" throughout the code, even though we decided not to handle species/subspecies in any special way; we're just searching by a taxonomy ID

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Taxonomy.java
```diff
@@ -76,16 +105,86 @@ public long getTaxonomyId() {
...
+
+    public Taxonomy getTaxonomyClass() {
+        return taxonomyClass;
+    }
+
+    public void setTaxonomyClass(Taxonomy taxonomyClass) {
+        this.taxonomyClass = taxonomyClass;
+    }
+
+    public enum Rank {SUBSPECIES, SPECIES, GENUS, ORDER, CLASS, UNKNOWN}
```
tskir: If I remember correctly we decided to not have an enum and to just store rank in plain text as represented in NCBI taxonomy tree

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/TaxonomySteps.java
tskir: OK, so I remember that we decided not to implement automatic construction of species/genus/order/class in _this_ ticket, but we should at least ensure that a ticket is created in the backlog for this

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/SwaggerConfig.java
```diff
@@ -92,7 +92,8 @@ public Docket metadataApi() {
                         new Tag("Sample Entity", "Sample metadata"),
                         new Tag("Study Entity", "Study metadata"),
                         new Tag("Taxonomy Entity", "Taxonomy definitions"),
-                        new Tag("WebResource Entity", "Study resources that can be accessed through web protocols")
+                        new Tag("WebResource Entity", "Study resources that can be accessed through web protocols"),
+                        new Tag("TaxonomyTree Entity", "Taxonomy tree")
```
tskir: If I remember correctly we decided to not use TaxonomyTree at all, storing everything directly in Taxonomy entity as well; why are there still references to TaxonomyTree then?

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Taxonomy.java
```diff
@@ -76,16 +105,86 @@ public long getTaxonomyId() {
...
+
+    @ApiModelProperty(position = 9, readOnly = true)
+    public Long getTaxOrderId() {
+        return (taxonomyOrder == null) ? -1 : taxonomyOrder.getTaxonomyId();
+    }
+
+    @ApiModelProperty(position = 11, readOnly = true)
+    public Long getTaxonomyClassId() {
+        return (taxonomyClass == null) ? -1 : taxonomyClass.getTaxonomyId();
+    }
```
tskir: Are these properties required? They are not intended to be used directly anyway, but are just for searching purposes instead. I. e. there is no scenario where someone would want to retrieve class ID **directly** from the Taxonomy entity. If I understand correctly, search by class/order/etc. is implemented by a column join, so it should still work without those entities, right?

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/TaxonomyTreeService.java
```diff
@@ -0,0 +1,28 @@
...
+
+package uk.ac.ebi.ampt2d.metadata.persistence.services;
+
+import java.util.List;
+
+public interface TaxonomyTreeService {
+
+    List<Long> findAllSpeciesAndSubspeciesTaxonomyIdsInATaxonomyTreeByTaxonomyId(long taxonomyId);
+
+    List<Long> findAllSpeciesAndSubspeciesTaxonomyIdsInATaxonomyTreeByTaxonomyName(String taxonomyName);
```
tskir: Again, why do we still have `TaxonomyTreeService` if we decided to get rid of TaxonomyTree entirely?

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Taxonomy.java
selvaebi: yes it does wordk , i have just put to show the user the taxonomyId of the parents directly instead of traversing through the url , this is not in our usecase but thought it would be handy for viewing

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/SwaggerConfig.java
selvaebi: sorry will remove that. i removed the entities but forgot to remove the references.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/TaxonomyTree.java
tskir: OK, then we'll need to create a ticket in the backlog for it

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Taxonomy.java
selvaebi: yes, we thought that, but anyways this enum can be a standard for searching the taxonomy by rank, the empty ranks will put in to UNKNOWN.

tskir: We don't want the empty rank to be "UNKNOWN", it should represent exactly what the NCBI taxonomy says it to be (be it "species", "subspecies", "strain" or "no rank"). In theory we could create a huge ENUM with all possible NCBI taxonomy ranks, but it would be a pain to constantly maintain and synchronise it, so frankly I don't see the point. Just storing rank as a string is totally fine.

tskir: For the purposes of search by class/order/etc. it also won't be a problem because references to fixed ranks are stored in the appropriate columns already; i. e. we don't need to know that a certain taxonomy is of RANK.CLASS rank to be able to query by class, because it is already referenced to by the "class" column

selvaebi: okies then I can remove it

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Taxonomy.java
tskir: But then there's additional information to be represented for easier viewing, for example names of genus/order/class etc. would also be convenient. The problem is that it creates lots of code we have to maintain; e. g. if in the future we decide to add an additional level for fixed-rank taxonomies, we will have to update this part as well. Since the API is intended for programmatic access, the ease of viewing isn't the priority

selvaebi: ya, I understand, then it would be easier for me to delete this extra code then :)

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/TaxonomyTreeService.java
tskir: Yes, but since we do not have a `TaxonomyTree`, surely we should not have such thing as a `TaxonomyTreeService`; at least call it `TaxonomyService`

metadata-ws/src/test/java/uk/ac/ebi/ampt2d/metadata/cucumber/TaxonomySteps.java
selvaebi: yes the load taxonomy tree ticket can be used for auto load of taxonomytree

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/TaxonomyTree.java
tskir: Ticket created https://www.ebi.ac.uk/panda/jira/browse/T2D-349, resolving discussion

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Taxonomy.java
```diff
@@ -54,11 +56,33 @@
...
     @Size(max = 255, min = 1)
+    @Column(unique = true)
     private String name;
 
-    @ManyToMany
+    @ApiModelProperty(position = 4)
     @JsonProperty
-    private List<Taxonomy> ancestors;
+    @Column
+    private String rank;
```
tskir: I think that we should still have a `@NotNull` condition on the `rank` property. When NCBI specifies that a taxon is of "no rank", this doesn't really have any special meaning which we would need to process; we can just treat all ranks equally and, in this case, simply store the "no rank" literal in this property. However, it must always be specified to contain _something,_ hence we need the `@NotNull` annotation

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/TaxonomyServiceImpl.java
```diff
@@ -0,0 +1,80 @@
...
+public class TaxonomyServiceImpl implements TaxonomyService {
+
+    @Autowired
+    private TaxonomyRepository taxonomyRepository;
+
+    @Autowired
+    private EntityManager entityManager;
+
+    @Override
+    public List<Long> findAllSpeciesAndSubspeciesTaxonomyIdsInATaxonomyTreeByTaxonomyId(long taxonomyId) {
```
tskir: Name of this method is very long and still mentions species and subspecies; let's rename it to something like 'findAllTaxonomiesByTaxonomyId' (or similar). Same for the second method

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/TaxonomyServiceImpl.java
```diff
@@ -0,0 +1,80 @@
...
+        List<Taxonomy> taxonomies = (List<Taxonomy>) jpaQueryFactory.from(qTaxonomy)
+                .leftJoin(qTaxonomy.taxonomySpecies, qTaxonomySpecies)
+                .leftJoin(qTaxonomy.taxonomyGenus, qTaxonomyGenus)
+                .leftJoin(qTaxonomy.taxonomyOrder, qTaxonomyOrder)
+                .leftJoin(qTaxonomy.taxonomyClass, qTaxonomyClass)
+                .where(qTaxonomySpecies.taxonomyId.eq(taxonomyId)
+                        .or(qTaxonomyOrder.taxonomyId.eq(taxonomyId)
+                                .or(qTaxonomyGenus.taxonomyId.eq(taxonomyId))
+                                .or(qTaxonomyClass.taxonomyId.eq(taxonomyId)))
+                        .or(qTaxonomy.taxonomyId.eq(taxonomyId))).fetch();
```
tskir: The genus and class comparisons are indented one way further than the other ones; lines 55-58 should be indented to the same level

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/TaxonomyServiceImpl.java
```diff
@@ -0,0 +1,80 @@
...
+            return new ArrayList<>();
+        }
+        return findAllSpeciesAndSubspeciesTaxonomyIdsInATaxonomyTreeByTaxonomyId(taxonomy.getTaxonomyId());
+    }
+
+    public List<Long> getSpeciesAndSubspeciesTaxonomyIds(List<Taxonomy> taxonomies) {
+        return taxonomies.parallelStream().filter(taxonomy -> {
+            String rank = taxonomy.getRank();
+            if (rank.equals("class") || rank.equals("order") || rank.equals("genus")) {
+                return false;
```
tskir: I don't think there is need to perform this filtering at all, since entities such as samples and reference sequences won't ever contain these higher-rank taxonomies, and if _for some reason_ they do, then this needs to be displayed as well. Besides, this filtering requires additional computational time. It's simpler to just return all taxonomy IDs and then query  the required entities using them all

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/TaxonomyServiceImpl.java
```diff
@@ -0,0 +1,80 @@
...
+                .leftJoin(qTaxonomy.taxonomyGenus, qTaxonomyGenus)
+                .leftJoin(qTaxonomy.taxonomyOrder, qTaxonomyOrder)
+                .leftJoin(qTaxonomy.taxonomyClass, qTaxonomyClass)
+                .where(qTaxonomySpecies.taxonomyId.eq(taxonomyId)
+                        .or(qTaxonomyOrder.taxonomyId.eq(taxonomyId)
+                                .or(qTaxonomyGenus.taxonomyId.eq(taxonomyId))
+                                .or(qTaxonomyClass.taxonomyId.eq(taxonomyId)))
+                        .or(qTaxonomy.taxonomyId.eq(taxonomyId))).fetch();
+        return getSpeciesAndSubspeciesTaxonomyIds(taxonomies);
+    }
```
tskir: I think the current search algorithm could be improved. For example, imagine you want to filter studies by order. Under the current implementation, it does a query against **all columns** (qTaxonomy, qTaxonomySpecies, qTaxonomyGenus, ...), then gets the list of suitable taxonomy IDs including lower-level, _then_ queries the taxonomy repository with the list of these IDs. Very time-consuming. While in reality you could filter by the `taxonomyOrder` column directly using only a single join, because the order will be specified in there directly. So I think it would make sense, when querying via some taxonomy ID, to first get the rank of this ID, then query via an suitable column if it's available, or via the direct comparison if not.

tskir: I won't insist for this to be implemented in this ticket because it would break tests. But in the next ticket (loading taxonomy tree) it would be desirable

metadata-ws/src/test/resources/features/sample.feature
```diff
@@ -101,12 +106,12 @@ Feature: sample object
...
     And the response should contain 2 taxonomies
-    And the href of the taxonomy of taxonomies has items TAXONOMY_1,TAXONOMY_2
+    And the href of the taxonomy of taxonomies has items TAXONOMY_MUS_MUSCULUS,TAXONOMY_HOMO_SAPIEN
 
-    When I request DELETE for the taxonomies of TAXONOMY_1 of the SAMPLE
+    When I request DELETE for the taxonomies of TAXONOMY_MUS_MUSCULUS of the SAMPLE
     Then the response code should be 2xx
 
-    When I request DELETE for the taxonomies of TAXONOMY_2 of the SAMPLE
+    When I request DELETE for the taxonomies of TAXONOMY_HOMO_SAPIEN of the SAMPLE
```
tskir: I'm guessing it should've been `TAXONOMY_HOMO_SAPIENS`

metadata-ws/src/test/resources/features/study.feature
```diff
@@ -688,7 +691,7 @@ Feature: study object
     Examples:
       | base          | query             | url    |
       | taxonomy-id   | id=9606           | STUDY2 |
-      | taxonomy-name | name=Homo sapiens | STUDY2 |
+      | taxonomy-name | name=Homo Sapiens | STUDY2 |
```
tskir: It was actually correct the first time around, because species name is written starting with a lowercase letter
https://en.wikipedia.org/wiki/Binomial_nomenclature

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/services/TaxonomyServiceImpl.java
selvaebi: it is good you are right, I have implemented in this ticket itself and it works fine.

tskir: Great, thanks!

PR #73, opened 2019-09-21
evaluating liquibase

iyangar:


metadata-ws/src/main/resources/db/changelog/db.changelog-1.0.0.xml
```diff
@@ -0,0 +1,339 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
```
selvaebi: how is this xml generated ?

metadata-ws/src/main/resources/application.properties
```diff
@@ -50,3 +50,7 @@ swagger.apiinfo.termsOfServiceUrl=
 security.enabled=true
 security.oauth2.resource.filter-order=3
 security.oauth2.resource.jwt.key-value=-----BEGIN PUBLIC KEY-----\n@metadata-jwtKey@\n-----END PUBLIC KEY-----
+
+spring.datasource.driver-class-name=org.postgresql.Driver
```
selvaebi: why do we need these values , if we already have spring.jpa.database-platform

metadata-load/src/test/resources/application.properties
```diff
@@ -20,3 +20,4 @@ spring.jmx.enabled=false
 entrez.api.key=
 entrez.api.attempts=10
 entrez.api.delay=1000
+liquibase.enabled=false
```
selvaebi: is it mandatory in metadata-load as well

metadata-ws/src/main/resources/db/changelog/db.changelog-1.0.0.xml
iyangar: Liquibase has a command generateChangeLog to get the changes from the existing DB. The same is used to get changelog from existing master build.

metadata-ws/src/main/resources/application.properties
iyangar: Liquibase documentation has this. I can verify if it works without above value.

metadata-ws/src/main/resources/db/changelog/db.changelog-1.0.0.xml
selvaebi: so you need to add it to the build phase to automatically generate it

metadata-load/src/test/resources/application.properties
iyangar: Verified removing this, test fails

metadata-ws/src/main/resources/application.properties
iyangar: removed and verified

metadata-load/src/test/resources/application.properties
tskir: Sure, we don't want to enable Liquibase in the test environment, only in the main one. Marking this as resolved.

metadata-ws/src/main/resources/db/changelog/db.changelog-1.0.0.xml
```diff
@@ -0,0 +1,339 @@
...
+		<addForeignKeyConstraint baseColumnNames="analysis_id" baseTableName="analysis_files" constraintName="fkq67f06oxvmdj36hdu363cpg1e" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="analysis" validate="true"/>
+	</changeSet>
+	<changeSet author="metauser (generated)" id="1569329013185-52">
+		<addForeignKeyConstraint baseColumnNames="study_id" baseTableName="study_child_studies" constraintName="fktfwfp9t1op8ku2pcjrw8c80nw" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="study" validate="true"/>
+	</changeSet>
+
+	<changeSet id="tag-1.0" author="metauser">
+		<tagDatabase tag="1.0" />
+	</changeSet>
+</databaseChangeLog>
```
tskir: Please add a file to the `docs/` folder explaining:
- How Liquibase works for our project (briefly)
- How to generate the new changelog for newly released version
- How to update the main `liquibase-changelog.xml`

iyangar: I feel confluence page may be a good choice here.

PR #74, opened 2019-09-23
T2D-216 Allow admins and object owners to access them before release date

selvaebi:


metadata-ws/src/test/resources/features/release-date.feature
```diff
@@ -184,3 +187,110 @@ Feature: Transitive release date control for child objects of a Study
...
+    When I request PATCH STUDY with patch and day tomorrow
+    Then the response code should be 2xx
+    When I request PATCH STUDY1 with patch and day tomorrow
+    Then the response code should be 2xx
+
+    When I set authorization with testuser having default role
+    And I request GET with value of STUDY
+    Then the response code should be 200
+    And I request GET with value of STUDY1
+    Then the response code should be 4xx
```
iyangar: It would be good to put exact HTTP code, especially in fail cases. Fox ex, response here could have been 404 or 401, for which the expectation can not be same.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/ReleaseDateAspect.java
```diff
@@ -72,13 +81,38 @@ public Object filterOnReleaseDateAdviceFindOne(ProceedingJoinPoint proceedingJoi
...
+        }
+
+        //Allow admins to access before release date
+        if (authentication.getAuthorities().stream()
+                .anyMatch(authority -> authority.getAuthority()
+                        .equals("ROLE_" + EnableSecurityConfig.ROLE_SERVICE_OPERATOR))) {
+            return result;
+        }
+
+        String commaSeperatedPermittedStudiesToTheUsers = ((CustomUsernamePasswordAuthenticationToken)
```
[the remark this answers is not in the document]
selvaebi: it is kind of self-documenting variable name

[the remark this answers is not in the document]
iyangar: The method name is getObjectUnlessBeforeReleaseDate and documentation has " If the returned
 object" which are generic. If you want to make Study specific please mention it somewhere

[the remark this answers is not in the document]
tskir: It would be sensible to rename to `commaSeparatedPermittedStudies` (since we're working in the context of a specific user anyway, and this is always implied). Also note the misspelling: _seperated_ → _separated_

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/ReleaseDateAspect.java
```diff
@@ -72,13 +81,38 @@ public Object filterOnReleaseDateAdviceFindOne(ProceedingJoinPoint proceedingJoi
...
+                ((OAuth2Authentication) authentication).getUserAuthentication()).getStudies();
+        List<String> permittedStudies = (commaSeperatedPermittedStudiesToTheUsers.isEmpty()) ? new ArrayList<>() : Arrays.asList
+                (commaSeperatedPermittedStudiesToTheUsers.split(","));
```
tskir: This string is long and bulky; would be better to rewrite it as a simple if-else statement (not the ternary if)

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/File.java
```diff
@@ -86,7 +86,31 @@
...
+            "INNER JOIN analysis on analysis_files.analysis_id = analysis.id " +
+            "INNER JOIN study on analysis.study_id = study.id " +
+            "WHERE file.id=id)")
+    @JsonIgnore
+    private LocalDate releaseDate;
+
+    /**
+     * Get the studyIds.
+     */
+
```
tskir: Extra empty line is present here; I would also suggest to expand description a bit, e. g., "Get the IDs of the studies which link to this object (used for access control)”, something like that

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/File.java
```diff
@@ -86,7 +86,31 @@
...
+    @Formula("(SELECT string_agg(concat(study.accession,'.',study.version),',') FROM file " +
+            "INNER JOIN analysis_files on file.id = analysis_files.files_id " +
+            "INNER JOIN analysis on analysis_files.analysis_id = analysis.id " +
+            "INNER JOIN study on analysis.study_id = study.id " +
+            "WHERE file.id=id)")
```
tskir: I would very much suggest to move the constant part of the expression (starting from `FROM` until the end of expression) into a string constant. So that the formula looks like something like this:

```java
@Formula("(SELECT string_agg(concat(study.accession,'.',study.version),',')" + fileQueryExpression + ")")
```

And also do the same for all other objects. It would _greatly_ reduce the possibility of errors when updating some of it in the future.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Publication.java
```diff
@@ -66,8 +68,21 @@ public Long getId() {
     @JsonIgnore
     private LocalDate releaseDate;
 
+    /**
+     * Get the studyIds.
+     */
+    @Formula("(SELECT string_agg(concat(study.accession,'.',study.version),',') FROM study_publications " +
```
tskir: Also, do we really want to have such a high granularity of control as to control studies based on their accession **and version?** I think just accession should be enough. For example imagine someone has access to a number of studies and then decides to roll out an updated version for several of them. This would necessitate updating all of the fields in Keycloak for every user which had access to those studies. Just access based on accession should be enough.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Taxonomy.java
```diff
@@ -178,4 +178,9 @@ public void setTaxonomyForRank(Taxonomy taxonomy, String rank) {
                 break;
         }
     }
+
+    @Override
+    public String getStudyIds() {
+        return null;
+    }
```
tskir: Could you please also document all cases where `null` is returned with something like “Access to entities of this kind is not restricted, hence the special value of `null` is returned”

metadata-ws/src/test/resources/features/release-date.feature
```diff
@@ -118,12 +118,13 @@ Feature: Transitive release date control for child objects of a Study
 
     # Set release dates for studies according to the scenario
     When I request PATCH STUDY1 with patch and day <S1_RELEASE>
-    Then the response code should be 200
+    Then the response code should be 2xx
```
tskir: As Sreenidhi mentioned elsewhere, it is better to check for specific status codes rather than for the general ones. Is there any reason to change from `200` to `2xx`?

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/ReleaseDateAspect.java
```diff
@@ -72,13 +81,38 @@ public Object filterOnReleaseDateAdviceFindOne(ProceedingJoinPoint proceedingJoi
...
+        String studiesAssociatedToEntities = ((Auditable) result).getStudyIds();
+        if (permittedStudies.stream()
+                .anyMatch(permittedStudy -> studiesAssociatedToEntities.contains(permittedStudy))) {
```
tskir: OK, so this way of doing it with a comma-separated list of studies is acceptable; but is there definitely no way to specify studies as separate resources in Keycloak? It would be more user friendly to whomever is going to be ultimaterly responsible for managing the user permissions

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Publication.java
selvaebi: okay that seems reasonable

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/ReleaseDateAspect.java
selvaebi: the feature which we are using in Keycloak for this functionality is kind of key-value pairs, so key being the userId values is kinda made a comma-separated list of study accessions.

metadata-ws/src/test/resources/features/release-date.feature
selvaebi: 200 and 204 are just successful with ok and no content so I wanted to show that it is successful.  but 404 and 403 might be surely needed to be differntiate as Sreenidhi mentioned.

tskir: OK, good point, agreed

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/ReleaseDateAspect.java
tskir: OK, understood. Anyway, the current implementation is very reasonable so no point of trying to optimise is further (especially considering EGA use case is far from deployment at this point)

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/ReleaseDateAspect.java
```diff
@@ -72,13 +80,46 @@ public Object filterOnReleaseDateAdviceFindOne(ProceedingJoinPoint proceedingJoi
...
+        }
+        permittedStudies = Arrays.asList(commaSeparatedPermittedStudies.split(","));
+        String studiesAssociatedToEntities = ((Auditable) result).getStudyIds();
+        if (permittedStudies.stream().anyMatch(permittedStudy ->
+                /*
+                * Below regex checks whether the permitted study is contained in the beginning
+                * or in between or at the end of the comma-selist of studiesAssociatedToEntities
+                */
+                studiesAssociatedToEntities.matches("(^" + permittedStudy + "$)|(^" + permittedStudy +
+                        ",)|(," + permittedStudy + ",)|(," + permittedStudy + "$)"))) {
```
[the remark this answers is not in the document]
tskir: Also, the way object ownership works should be documented (it's OK if very briefly) in the `docs/`. It's sufficient to just mention that a Keycloak server should be configured and the comma-separated list of study accessions (without a version) should be specified in a certain key-value pair for the user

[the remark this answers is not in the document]
selvaebi: will there be a comma in study accession ??

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/ReleaseDateAspect.java
selvaebi: Sorry, I explained it wrong but it is the same case just the key is a studies which will be included in the jwt token.

docs/release-date.md
```diff
@@ -16,3 +16,12 @@ To enforce that, transitive release date control is implemented. It works in the
   + If at least one released Study links to an entity (including transitive linking, e. g. Study linking to Analysis
     linking to File), this entity is considered to be released and becomes retrievable through the API.
   + Otherwise, the entity is not released and not retrievable through the API.
+  
+## Access to Entities before release date
+
+In certain cases the studies and its related entities should be accessible before release date for owners and 
+journals. And by default admins will be able to access the entities regardless of the release date.
```
tskir: What (or who) exactly are journals? I haven't heard of this term before in this context

docs/release-date.md
```diff
@@ -16,3 +16,12 @@ To enforce that, transitive release date control is implemented. It works in the
...
+
+* To make a study accessible before release date for any user (owner or journals) , corresponding study accessions
```
tskir:
```suggestion
* To make a study accessible before release date for any user (owner or journals), corresponding study accessions 
```

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/aop/ReleaseDateAspect.java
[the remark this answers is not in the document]
tskir: Right, so about this one: while I agree it's unlikely, we still shouldn't allow for a comma to be contained in the accession. I created a ticket in the backlog: https://www.ebi.ac.uk/panda/jira/browse/T2D-363 (so we won't be doing it as part of this PR).

tskir:
And also, something appears to be wrong with the Jenkins build, could you please check? Travis is OK

tskir:
Some notes:
* Applied changes to docs wording myself in a separate commit to avoid having this PR unmerged for a week+ with some only minor changes
* Restarted Jenkins build, everything's OK

PR #75, opened 2019-09-26
T2D-279 moved to default env variable for build directory

selvaebi:


selvaebi:
Sorry, the buildDirectory is used for Jenkins deployment file to differentiate different environments.

PR #79, opened 2019-10-14
Add retry for ENA XML API

iyangar:
Unit testing is done by simulating HTTP 500 error with SraObjectsImporterThroughApi->getXml()

### DIFF
--- a/metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/api/SraObjectsImporterThroughApi.java
+++ b/metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/api/SraObjectsImporterThroughApi.java
@@ -19,6 +19,11 @@
 package uk.ac.ebi.ampt2d.metadata.importer.api;
 
 import org.springframework.core.convert.converter.Converter;
+import org.springframework.http.HttpMethod;
+import org.springframework.retry.annotation.Backoff;
+import org.springframework.retry.annotation.Retryable;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.web.client.RestTemplate;
 import uk.ac.ebi.ampt2d.metadata.importer.ObjectsImporter;
 import uk.ac.ebi.ampt2d.metadata.importer.extractor.FileExtractorFromAnalysis;
 import uk.ac.ebi.ampt2d.metadata.importer.extractor.PublicationExtractorFromStudy;
@@ -157,5 +162,25 @@ public class SraObjectsImporterThroughApi extends ObjectsImporter {
                 });
 
         return analysisAccessions;
+    }
+
+    @Override
+    @Transactional(rollbackFor = Exception.class)
+    public Study importStudy(String accession) throws Exception {
+        String xml = getXml(accession);
+        StudyType studyType = sraStudyXmlParser.parseXml(xml, accession);
+        Study study = studyConverter.convert(studyType);
+        StudyType.STUDYLINKS studylinks = studyType.getSTUDYLINKS();
+        study.setPublications(publicationExtractorFromStudy.getPublications(studylinks));
+        study.setResources(webResourceExtractorFromStudy.getWebResources(studylinks));
+        study = extractAnalysisFromStudy(studyType, study);
+        return study;
+    }
+
+    @Retryable(maxAttemptsExpression="#{${api.attempts}}",
+            backoff=@Backoff(delayExpression="#{${api.delay}}"))
+    private String getXml(String accession) throws Exception {
+        RestTemplate restTemplate = new RestTemplate();
+        return restTemplate.exchange(SraXmlRetrieverThroughApi.ENA_API_URL, HttpMethod.GET, null, String.class, accession).getBody();
     }
 }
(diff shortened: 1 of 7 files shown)


metadata-load/src/test/resources/application.properties
```diff
@@ -15,8 +15,9 @@ metadata.datasource.driver-class-name=org.h2.Driver
...
 spring.jmx.enabled=false
 
-# Entrez API querying settings.
-# For testing purposes, using no API key and increasing delay to 1000 to accommodate lower allowed request rate
+# Entrez API key
 entrez.api.key=
-entrez.api.attempts=10
-entrez.api.delay=1000
+
+# Entrez and ENA API retry attempts and request delay (can be as low as 500 ms).
```
selvaebi: it is said as delay can be as low as 500 ms, but the delay is 1000 ??

iyangar: For testing purpose delay is kept as 1000 for better accommodation

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/api/SraObjectsImporterThroughApi.java
```diff
@@ -158,4 +163,24 @@ protected Analysis extractStudyFromAnalysis(AnalysisType analysisType, Analysis
 
         return analysisAccessions;
     }
+
+    @Override
+    @Transactional(rollbackFor = Exception.class)
+    public Study importStudy(String accession) throws Exception {
```
tskir: Is there a reason to duplicate code from `ObjectsImporter` and `SraXmlRetrieverThroughApi` here? It seems like it would be sufficient to just add a `@Retryable` annotation to `SraXmlRetrieverThroughApi.getXml()`

iyangar: Spring will not be able to resolve. Below error will be thrown
"Bean named 'sraXmlRetrieverThroughApi' is expected to be of type 'uk.ac.ebi.ampt2d.metadata.importer.api.SraXmlRetrieverThroughApi' but was actually of type 'com.sun.proxy.$Proxy137'"

tskir: This appears to be because we're relying on a specific implementation (`SraXmlRetrieverThroughApi`) rather than the interface (`SraXmlRetrieverByAccession`). It looks like this can can be fixed by relying on an interface (which also makes more sense). See here, for example: https://stackoverflow.com/questions/14509142/beannotofrequiredtypeexception-due-to-autowired-fields

In either case, code duplication is not a good option

metadata-load/src/main/resources/application.properties
```diff
@@ -30,3 +30,7 @@ spring.main.web-environment=false
 entrez.api.key=@entrez.api.key@
 entrez.api.attempts=10
 entrez.api.delay=500
+
+# ENA API settings to accommodate slow response
```
tskir: This needs to be slightly rephrased: what we're fixing here is not really a slow response, but _rare and sporadic errors_

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/api/SraObjectsImporterThroughApi.java
iyangar: Thanks, ObjectImporter->importStudy() has sraXmlRetrieverByAccession.getXml(). Did i miss anything

tskir: For example, see `MetadataImporterMainApplicationConfiguration`, where the autowiring actually happens. This is where we autowire by specific class (`SraXmlRetrieverThroughApi sraXmlRetrieverThroughApi`) rather than by an interface.

I think something like this should work: https://stackoverflow.com/a/31138957

tskir: I. e. make it like `SraXmlRetrieverByAccession sraXmlRetrieverThroughApi`

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/configuration/MetadataImporterMainApplicationConfiguration.java
```diff
@@ -66,7 +67,7 @@ public ReferenceSequenceXmlRetrieverThroughEntrezApi assemblyXmlRetrieverThrough
 
     @Bean
     @ConditionalOnProperty(name = "import.source", havingValue = "API")
-    public ObjectsImporter objectImporterThroughEnaApi(SraXmlRetrieverThroughApi sraXmlRetrieverThroughApi,
+    public ObjectsImporter objectImporterThroughEnaApi(SraXmlRetrieverByAccession sraXmlRetrieverByAccession,
```
tskir: Actually I think that here you need to make it `SraXmlRetrieverByAccession sraXmlRetrieverThroughApi`. So you depend on an interface (`SraXmlRetrieverByAccession`), but name a variable after a specific class (`sraXmlRetrieverThroughApi`). This way, Spring will know which specific bean to inject. Currently it highlights as a multiple autowire error:
a

PR #80, opened 2019-10-21
T2D-353 — Fix bug with discordant taxonomies in the reference sequences

iyangar:


metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/events/TaxonomyEventHandler.java
```diff
@@ -77,6 +77,9 @@ public Taxonomy importTaxonomyTree(Taxonomy taxonomy) throws Exception {
                 long taxIdLong = Long.parseLong(taxId);
                 Taxonomy taxonomyParent = importTaxonomyTree(new Taxonomy(taxIdLong));
                 taxonomy.setTaxonomyForRank(taxonomyParent, rankEnum.toString());
+            } else if (taxonomy.getRank().equals(RANK.SPECIES.name) && taxonomy.getTaxonomySpecies() == null) {
```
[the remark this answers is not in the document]
iyangar: I think i missed comparing `rankEnum` with RANK.SPECIES. if (rankEnum.equals(RANK.SPECIES) && taxonomy.getTaxonomySpecies() == null)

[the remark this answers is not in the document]
selvaebi: u check it outside for loop, check whether the rank is species and setTaxonomySpecies.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/events/TaxonomyEventHandler.java
```diff
@@ -77,6 +77,9 @@ public Taxonomy importTaxonomyTree(Taxonomy taxonomy) throws Exception {
...
+                // self-reference at species level
+                taxonomy.setTaxonomySpecies(taxonomy);
```
tskir: I think this can be simplified even more. The issue we're solving here affects not only species taxonomies; if we import an order taxonomy, its `taxonomyOrder` field will also be empty because of the same problem.

Instead, you could include a check inside the for-loop: if the `rank` is the same as `rankEnum` (meaning the taxonomy's rank is one of the fixed-rank list), then call `setTaxonomyForRank` with the `taxonomy` itself.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/events/TaxonomyEventHandler.java
```diff
@@ -77,6 +77,9 @@ public Taxonomy importTaxonomyTree(Taxonomy taxonomy) throws Exception {
...
             }
         }
```
tskir: Shouldn't this method include `return taxonomy` at the end? Or am I missing something?

iyangar: I think "return taxonomyRepository.save(taxonomy);" is at the end

tskir: You're right! I just missed it somehow

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/events/TaxonomyEventHandler.java
```diff
@@ -77,6 +77,9 @@ public Taxonomy importTaxonomyTree(Taxonomy taxonomy) throws Exception {
                 long taxIdLong = Long.parseLong(taxId);
                 Taxonomy taxonomyParent = importTaxonomyTree(new Taxonomy(taxIdLong));
                 taxonomy.setTaxonomyForRank(taxonomyParent, rankEnum.toString());
+            } else if (taxonomy.getRank().equals(RANK.SPECIES.name) && taxonomy.getTaxonomySpecies() == null) {
+                // self-reference at species level
```
tskir: About the data integrity violation exception: I have a feeling that you have to save an entity before you can self-reference it. So I suggest you rewrite it in such a way that the entity is first saved (with an empty `taxonomySpecies` field), and _then_ the `setTaxonomyForRank` is called.

tskir: (As discussed in Slack, this does not solve the problem of the data integrity violation.)

OK, so I've tested the current code (as it is in this PR right now), and I see that the actual self-referencing actually does work. For example, if you import a study ERP016904, you get the species taxonomy which correctly refers back to itself in `taxonomySpecies`. You can view it in the metadata web service with no problem. So, first of all, we don't need to save the taxonomy prior to self-referencing it; and secondly, the problem appears to be not in the self-referencing part at all.

The error thrown in the tests is for the study ERP000054. It doesn't matter if you import it alone or with other studies, it always throws an error:

```
SEVERE: Encountered Exception for study accession ERP000054
Nov 06, 2019 9:57:00 AM uk.ac.ebi.ampt2d.metadata.importer.MetadataImporterMainApplication lambda$run$1
SEVERE: An analysis must be mapped against one or more reference sequences
2019-11-06 09:57:52.279  WARN   --- [           main] o.h.engine.jdbc.spi.SqlExceptionHelper   : SQL Error: 90053, SQLState: 90053
2019-11-06 09:57:52.281 ERROR   --- [           main] o.h.engine.jdbc.spi.SqlExceptionHelper   : Scalar subquery contains more than one row; SQL statement:
select study0_.id as id1_9_0_, study0_.last_modified_date as last_mod2_9_0_, ...
[Long SQL statement redacted for brevity]
```

It looks like the code which checks the validity of an analysis is doing something wrong (in a way not compatible with self referencing). Since this part was recently changed, it seems likely. So I suggest you look into that part

tskir: Oh, wait a minute, the issue with ERP000054 is that its analysis https://www.ebi.ac.uk/ena/data/view/ERZ000001&display=xml indeed does not have a reference sequence, so this is *why* we're using at as an example of a study with an invalid analysis. So the problem is not that, I must have misread the logs

tskir:
Rebased on top of https://github.com/EBIvariation/metadata-ws/pull/81, let's see if it fixes the problem after all

.travis.yml
```diff
@@ -2,11 +2,19 @@ sudo: false
...
 
+git:
+  quiet: true
+
 jdk:
   - openjdk8
 
 install:
-  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -PTravis
+  - >
```
iyangar: Any reason for `>` character?

tskir: Thanks for pointing it out; I actually meant to use `|` instead. Fixed now.

Both of those symbols denote a block scalar literal in YAML; only `|` keeps newlines as is, and `>` is _supposed_ to replace them with spaces. So theoretically `>` _should not_ have worked. But apparently Travis has its own way of parsing `.travis.yml`, which is not entirely consistent with YAML specification.

Either way, now everything's fine

tskir:
(@selvaebi Sorry, requested review by mistake)

PR #83, opened 2019-12-09
T2D-359 — Syncronise list of technology types with SRA schema

tskir:
Synchronised list of supported technology types with SRA schema. Added documentation, both in comments and in the `docs/` folder.

PR #84, opened 2019-12-13
T2D-363 — Validate study accession on creation

tskir:
* Enforce validation of accession; don't allow comma or dot symbols
* Update documentation to clarify that comma is not accepted in accession
* Add a test case for not allowing invalid study accessions

### DIFF
--- a/docs/using-the-service.md
+++ b/docs/using-the-service.md
@@ -1,4 +1,8 @@
-# Querying the metadata service
+# Using the metadata service
+
+## Study queries
+
+Please note that a comma symbol (`,`) is not permitted to be used in a study accession, because this symbol is used to separate study accessions when configuring access control.
 
 ## Analysis queries
 
--- a/metadata-ws/src/test/resources/features/study.feature
+++ b/metadata-ws/src/test/resources/features/study.feature
@@ -1013,7 +1013,11 @@ Feature: study object
     Then the response code should be 400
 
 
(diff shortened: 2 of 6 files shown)


docs/using-the-service.md
```diff
@@ -0,0 +1,13 @@
+# Using the metadata service
+
+## Study queries
+
+Please note that a comma symbol (`,`) is not permitted to be used in a study accession, because this symbol is used to separate study accessions when configuring access control.
```
tcezard: We should document the dot as well.
```suggestion
Please note that a comma (`,`) and dot (.) symbols are not permitted to be used in a study accession. Dot is used to separate a study accession and version and comma is used to separate study accessions when configuring access control.
```

tskir:
Rebased the PR on top of master after the merge of #83

metadata-ws/src/test/resources/features/study.feature
```diff
@@ -1011,3 +1011,9 @@ Feature: study object
 
     When I request PATCH STUDY with content {""}
     Then the response code should be 400
+
+
+  Scenario: Trying to create a study with an invalid accession name (containing a comma symbol) must fail
```
andresfsilva: It would be good to include another one to check it fails when the accession has a `dot`

tskir: Thank you for a suggestion, addressed in https://github.com/EBIvariation/metadata-ws/pull/84/commits/c2233c1301e04bee15af6874c318e29ee1748874

PR #85, opened 2020-03-06
EVA-1805 — Handle studies or their components being suppressed in ENA

tskir:
* **Migrated to the new ENA XML API:** https://www.ebi.ac.uk/ena/data/view/ERP113859&display=xml → https://www.ebi.ac.uk/ena/browser/api/xml/ERP113859. Advantages of the new API:
  + It returns a 404 for entities which are suppressed or missing (old API always returned 200, and there was no reliable way to test for a status)
  + It is actively maintained, while the old API will be deprecated at some point
* **All entities for which ENA returns 404, indicating suppressed or missing status, are now being skipped without raising an exception.** (Although they are still logged for manual investigation purposes.)
* **Wrote documentation describing how suppressed and missing entities are being handled.**
* Technical changes
  + Removed unused `removeRootTagsFromXmlString` method (not needed by the new XML API)
  + Updated test files to reflect slight structural changes in the new XML API


metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -144,7 +144,9 @@ public ObjectsImporter(
 
     @Transactional(rollbackFor = Exception.class)
     public Study importStudy(String accession) throws Exception {
+        IMPORT_LOGGER.log(Level.INFO, "Importing study " + accession);
         String xml = sraXmlRetrieverByAccession.getXml(accession);
+        if (xml == null) { return null; }
```
jmmut: I guess it's probably too late to use `Optional` in this project consistently, but in the future consider using it instead of nullable return values.
If you didn't know about [`Optional`](https://mkyong.com/java8/java-8-optional-in-depth/), it forces syntactical checks, improving clarity and reducing the need to remember which values can be null, as it has to be handled explicitly.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/api/SraXmlRetrieverThroughApi.java
```diff
@@ -18,22 +18,41 @@
...
     private RestTemplate restTemplate = new RestTemplate();
 
-    @Retryable(maxAttemptsExpression="#{${ena.api.attempts}}",
-            backoff=@Backoff(delayExpression="#{${ena.api.delay}}"))
+    @Retryable(maxAttemptsExpression="#{${ena.api.attempts}}", backoff=@Backoff(delayExpression="#{${ena.api.delay}}"))
     @Override
     public String getXml(String accession) {
-        return restTemplate.exchange(ENA_API_URL, HttpMethod.GET, null, String.class, accession).getBody();
+        try {
+            return restTemplate.exchange(ENA_API_URL, HttpMethod.GET, null,
```
jmmut: for simple GET rest calls I think you can just use `restTemplate.get(ENA_API_URL)` which will use sensible default parameters.

README.md
```diff
@@ -8,13 +8,13 @@ Available documentation:
...
 
 
 ## Tests
 
 The profile for testing in a non-Travis environment requires an ENA database connection to be configured.
 
-Unfortunately the corresponding properties can't be populated using a Maven profile because they are defined inside
-the src/test/resources folder, so they must be explicitly set before running said tests.
\ No newline at end of file
+Unfortunately the corresponding properties can't be populated using a Maven profile because they are defined inside the src/test/resources folder, so they must be explicitly set before running said tests.
```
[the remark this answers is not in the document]
tskir: The suggestions you've left are really valuable! However, I'm afraid that if I start to dig deep into them, this will delay the ticket by several more days, and its scope will expand. I've created a separate ticket to implement the improvements you are suggesting: https://www.ebi.ac.uk/panda/jira/browse/EVA-1896

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
tskir: That's cool, I didn't know about `Optional`! I don't think it's too late to use it consistently. I've included it as part of the ticket mentioned above.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/api/SraXmlRetrieverThroughApi.java
tskir: I don't see a `get()` method in RestTemplate. Did you mean `getForEntity()`?

Otherwise, this makes sense and is indeed simpler. I replaced all `exchange()` calls with `getForEntity()`: https://github.com/EBIvariation/metadata-ws/pull/85/commits/668de0e5d718153f2ab3256b9992775e57861a61

jmmut: yes, I didn't have the doc at hand. There are 2 versions: getForEntity and getForObject. The first returns `ResponseEntity<T>` (`ResponseEntity<String>` in this case) to check http return codes and such. getForObject returns just `T` if you are interested in the content only. https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/client/RestTemplate.html

PR #86, opened 2020-03-13
EVA-1814 — Code updates after test run, version 5

tskir:
Quite a lot of small and medium changes addressing smaller issues discovered during the latest test run.

## Performance improvements
* Disabled `@Transactional` import for studies and analyses. Adjusted the tests accordingly.
* If the sample is already present, it will not be imported again, and no repeated API calls will be made.

Those two measures have sped up the import by approximately a factor of 10. It might not be obvious that improvement is this large, since previous iterations took about 15 hours and this one 8 hours. However, in previous iterations, a study with an especially huge number of samples was essentially skipped (due to a bug fixed in https://github.com/EBIvariation/metadata-ws/pull/85). Without those improvements, current test run iteration would have taken several days.

## Usability improvements
* Wrote documentation on configuring and running the pipeline, both on the cluster and on setting up local development environment. For the credentials, it includes pointers to the relevant pages in JIRA.
* Added template for `settings.xml` to be able to build the project with Oracle drivers.
* Input study accessions can now be read from an arbitrary text file, and it doesn't have to be in the `resources/` folder (not sure why it was implemented this way in the first place).
* In case an exception occurs during import, the error messages now include the full stack trace, which is much more helpful.

## Bug fixes
* Updated the `uk.ac.ebi.ena.sra.sra-xml` package from version 1.5.48 to 1.5.59. This was necessary because version 1.5.48 did not have support for an experiment type of “Target sequencing”, which is used in several EVA analyses.
* Updated literals for start and end tags in the XML returned by Entrez API to support the small changes which they made to their schema.
* Removed faulty validation for BioSamples accession field. Since we don't know the exact rule or regexp to validate those accessions, it's probably wisest not to do this at all and rely on ENA in providing this data.


metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/api/SraObjectsImporterThroughApi.java
```diff
@@ -108,6 +108,7 @@ protected Study extractAnalysisFromStudy(StudyType studyType, Study study) throw
             } catch (Exception exception) {
                 IMPORT_LOGGER.log(Level.SEVERE, "Encountered Exception for accession " + analysisAccession);
                 IMPORT_LOGGER.log(Level.SEVERE, exception.getMessage());
+                exception.getCause().printStackTrace();
```
jmmut: I recommend to print the stacktrace through the logger, to make sure it is close to the 2 previous messages.

settings.xml
```diff
@@ -0,0 +1,26 @@
+<settings>
+<servers>
+  <server>
+    <id>maven.oracle.com</id>
+    <username></username>
+    <password></password>
+    <configuration>
+      <basicAuthScope>
+        <host>ANY </host>
+        <port>ANY </port>
```
jmmut: I expect this should make a difference, but I'm not sure, as it looks like it worked for you as it is now.
```suggestion
        <host>ANY</host>
        <port>ANY</port>
```

docs/import/running-the-import.md
```diff
@@ -0,0 +1,76 @@
...
+sed -ie "s|<username></username>|<username>${ORACLE_USERNAME}</username>|" settings.xml
+sed -ie "s|<password></password>|<password>${ORACLE_PASSWORD}</password>|" settings.xml
+mvn -s settings.xml clean install -DskipTests=true
+```
+
+Write the list of studies to import to `studies-to-import.txt`.
+
+Run the import:
+```bash
+time bsub -M 20G -I java -Xmx18g \
```
sundarvenkata-EBI: Also add the -Rusage switch...

docs/import/running-the-import.md
```diff
@@ -0,0 +1,76 @@
...
+export BRANCH=master
+
+# Path where Maven is installed (or will be installed)
+# See also instructions below
+export MAVEN_INSTALL_PATH=
+export PATH=$MAVEN_INSTALL_PATH/bin:$PATH
+
+# Oracle credentials to download the database drivers
+# See Confluence → Variation home → IT infrastructure → External services credentials → Oracle
+export ORACLE_USERNAME=
```
sundarvenkata-EBI: Alright, so this is to login to the Oracle website. At first sight, I thought this was some Oracle database. Could this be renamed to ORACLE_SSO_USERNAME or something similar for clarity?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/MetadataImporterMainApplication.java
```diff
@@ -97,9 +97,8 @@ public void run(ApplicationArguments applicationArguments) throws Exception {
         String accessionFilePath = accessionsFilePath.get(0);
         Set<String> accessions;
         try {
-            accessions = new HashSet<>(Files.readAllLines(Paths.get(MetadataImporterMainApplication.class
-                    .getClassLoader().getResource(accessionFilePath).toURI())));
-        } catch (NullPointerException | URISyntaxException exception) {
+            accessions = new HashSet<>(Files.readAllLines(Paths.get(accessionFilePath)));
```
sundarvenkata-EBI: so, the rationale to simplify the code here is because we will always be using local files?

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -142,7 +142,6 @@ public ObjectsImporter(
         this.taxonomyEventHandler = taxonomyEventHandler;
     }
 
-    @Transactional(rollbackFor = Exception.class)
```
sundarvenkata-EBI: @tskir I am fine with making the trade-off  for performance. But, have you confirmed the original reasoning behind the transactional annotation (with the previous devs, perhaps)? Was that to prevent stale reads while the relevant Oracle table is being updated? If there are specific downsides to not having this as a transaction, could you please document it through comments/readme etc.,?

settings.xml
tskir: Wonder where those stray space symbols came from... Thanks, will fix

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/api/SraObjectsImporterThroughApi.java
tskir: I just didn't know how to do this using the logger. Turns out I can just do

```
IMPORT_LOGGER.log(Level.SEVERE, exception.getMessage(), exception);
```
And it will log both the message and the stack trace automatically.

tskir: Addressed in https://github.com/EBIvariation/metadata-ws/pull/86/commits/608ed7c7bf51ffab28ecd9637833f8cf5bf94477.

docs/import/running-the-import.md
tskir: Thanks, fixed in https://github.com/EBIvariation/metadata-ws/pull/86/commits/608ed7c7bf51ffab28ecd9637833f8cf5bf94477

docs/import/running-the-import.md
tskir: I agree, fixed in https://github.com/EBIvariation/metadata-ws/pull/86/commits/608ed7c7bf51ffab28ecd9637833f8cf5bf94477

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/MetadataImporterMainApplication.java
tskir: Exactly. Indeed, I don't know what was the rationale to define this as a resource file in the first place. The list of studies is always a simple local file supplied prior to running, not in any way part of the package

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
tskir: Luckily I already know the original reasoning because I was there when the decision was made. The idea was that if we make the import transactional, it will always result in a clean (but possibly incomplete) state in the target database, as any study or analysis with exceptions during import would be cleanly rolled back. This is not important for the EVA use case anyway, since even a single unexpected exception in the import process means that it has ultimately failed. I've added the explanation to documentation: https://github.com/EBIvariation/metadata-ws/pull/86/commits/608ed7c7bf51ffab28ecd9637833f8cf5bf94477

docs/import/running-the-import.md
```diff
@@ -49,16 +49,16 @@ Clone the repository and build the package:
...
+sed -ie "s|<password></password>|<password>${ORACLE_SSO_PASSWORD}</password>|" settings.xml
 mvn -s settings.xml clean install -DskipTests=true
 ```
 
 Write the list of studies to import to `studies-to-import.txt`.
 
 Run the import:
 ```bash
-time bsub -M 20G -I java -Xmx18g \
+time bsub -M 20G -I java -Xmx18g -Rusage \
```
sundarvenkata-EBI: Sorry I wasn't clear earlier... Rusage argument is for bsub as in -R "rusage[mem=20480]"

docs/import/missing-and-suppressed-entities.md
```diff
@@ -28,3 +28,9 @@ The skipped entities are being logged with a `WARN` type and a specitic error me
 ```
 2020-03-06 11:11:25.031  WARN 29784 --- [           main] u.a.e.a.m.importer.ObjectsImporter       : Accession not found in ENA ERS3377874
 ```
+
+# Handling exceptions
+
+During the earlier stages of development, import of Study and Analysis entities was defined to be `@Transactional`, meaning that if any error occurs during import, all changes to the database would be rolled back. This would guarantee a clean but possibly incomplete state after import, where studies and analyses would either be imported completely, or not imported at all.
+
+However, test runs demonstrated that doing a transactional import decreases performance by at least a factor of 10, which is critical. For EVA use case, since the goal is to import all data for all studies, even a single unhandled exception during import means that the import as a whole was a failure and hence the results are unusable in production. Hence, there is no reason to do a transactional import anyway.
```
sundarvenkata-EBI: So if a specific import fails, how is the data rolled back in production?

docs/import/running-the-import.md
tskir: Oops, sorry about that! `-Rusage` totally seemed like a flag which Java might have, so I didn't even check it :) Fixed in https://github.com/EBIvariation/metadata-ws/pull/86/commits/3f3755f4b2643a1f0190e53179842da217cbf607.

docs/import/missing-and-suppressed-entities.md
tskir: You keep asking good questions! Yes, if the import fails, there's currently no way to roll it back. One possible solution here would be to use two databases and to run the import on one while another one serves as production, and to alternate between them. It may be an ugly hack, but for now we don't really have any other way to proceed—if we do it in transactions, the import will just take too much time to be reasonably used. I have amended the documentation about this in https://github.com/EBIvariation/metadata-ws/pull/86/commits/3f3755f4b2643a1f0190e53179842da217cbf607.

tskir: I agree that this is not optimal; and maybe in the future we can spend more time on this and either improve performance of the import pipeline, or design some other solution to this problem.

docs/setting-up-environment.md
```diff
@@ -0,0 +1,85 @@
...
+In IDEA, open File → Project Structure. Change the “Project SDK” to Java 8.
+
+## 2. Configure Maven
+* Open File → Settings → Build, Execution, Deployment → Build Tools → Maven and change the location of the “User settings file” to `settings.xml` in the root of the repository. To change the setting, you will need to enable the “Override” checkbox on the right.
+* In the `settings.xml` file, fill in `ORACLE_SSO_USERNAME` and `ORACLE_SSO_PASSWORD` (see above).
+* In IDEA, right click on the project → Maven → Reimport. Wait for the process to complete.
+* Right click on the project → Maven → Generate sources and update folders.
+
+## 3. Create and configure local Postgres database
+If necessary, install and enable Postgres installation:
```
andresfsilva, line 24: It would be good to give the option to use docker as well with some command like
```
docker run --name postgres -p 5432:5432 -e POSTGRES_PASSWORD=pass -d postgres
```

tskir: Thanks, a good point! Addressed in https://github.com/EBIvariation/metadata-ws/pull/86/commits/8b5ca5ac8341bc5c655f664fe2776dfd67037331.

PR #87, opened 2020-06-02
EVA-1980 Add support for import base on project accessions

jmmut:
## Notes

- Added a One-to-One relationship between Study and Project
- When a study or a project is imported, the other one is imported too
- Analysis have both a link to a Study and a link to a Project
- The main integration test is SraObjectsImporterThroughApiTest#importProject

## remaining tasks 
- Update documentation: mention the script https://github.com/EBIvariation/configuration/blob/master/metadata-test-run-configuration.sh in docs/setting-up-environment.md
- Add an entry for project in src/main/java/uk/ac/ebi/ampt2d/metadata/SwaggerConfig.java and src/main/java/uk/ac/ebi/ampt2d/metadata/SpringDataRestConfig.java

## remaining questions:

- Do we need this class for projects too? src/main/java/uk/ac/ebi/ampt2d/metadata/aop/StudyDeprecationAspect.java
- Do we need the equivalent of StudyRepositoryCustom for project too? I added a simple ProjectRepository and I think the function defined in StudyRepositoryCustom is only used in the EGA import.
- The cucumber tests about study seem about creating new studies. Do we need those for projects too?
- I assumed projects are not relevant to be imported from the DB (for EGA) is that right? Maybe that has to be implemented for weird/supressed projects.


metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -142,23 +156,67 @@ public ObjectsImporter(
...
 
-    public Study importStudy(String accession) throws Exception {
-        IMPORT_LOGGER.log(Level.INFO, "Importing study " + accession);
+    public Project importProject(String accession) throws Exception {
+        LOGGER.info("Importing project " + accession);
+        ProjectType projectType = retrieveType(accession, sraProjectXmlParser);
+        if (projectType == null) { return null; }
+        Project project = convertProject(projectType);
+
+        String studyAccession = projectType.getIDENTIFIERS().getSECONDARYIDArray(0).getStringValue();
```
tskir, line 165: What worries me in that method is getSECONDARYID**Array.** Looks like it is possible for a project to have multiple studies (secondary IDs) linked to it after all. Sure, for the EVA use case it won't matter because it's always 1-to-1, but in general this doesn't appear to be the case

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
```diff
@@ -142,23 +156,67 @@ public ObjectsImporter(
...
+        StudyType studyType = retrieveType(studyAccession, sraStudyXmlParser);
+        if (studyType == null) {
+            throw new RuntimeException("All projects shold have a study ID");
+        }
+        Study study = convertStudy(studyType);
+        project = extractAnalysisFromProject(projectType, project, study);
```
tskir, line 171: If I understand this correctly (from the code & from the PR description), the implementation you made is versatile so that the analyses can be extracted both from the project and from the study. This is convenient but also kind of leads to code/logic duplication.

An alternative would be, for the EVA use case, to always start from a project a go strictly from there: a project imports _only_ study, which in turn is responsible for importing _all other_ entities it requires. For the EGA case, I'm not so sure how this would work. Maybe we could investigate the EGA use case in a separate ticket, to see what is the reason for the so much different import process. (Maybe we could just start always start from project in both use cases?)

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/PublicationExtractor.java
```diff
@@ -21,35 +21,51 @@
...
-            }
+    }
+
+    public List<Publication> getPublicationsFromStudy(StudyType.STUDYLINKS studylinks) {
+        if (studylinks == null) {
+            return new ArrayList<>();
+        } else {
+            return getPublicationsFromLinks(
+                    Stream.of(studylinks.getSTUDYLINKArray())
+                          .map(link -> link.getXREFLINK()));
```
tskir, line 60: This is another case where supporting importing child elements from both study and project seemingly leads to code duplication. Maybe we should only support one way, presumably the old one — extracting everything from study?

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Analysis.java
```diff
@@ -118,32 +118,37 @@
     @Column(nullable = false, columnDefinition = "TEXT")
     private String description;
 
-    @ApiModelProperty(position = 5, dataType = "java.lang.String", notes = "Url to a Study")
+    @ApiModelProperty(position = 5, dataType = "java.lang.String", notes = "Url to a Project")
+    @JsonProperty
+    @ManyToOne(optional = true) // TODO jmmut: I'm putting this as optional until we complete the project import from DB
```
tskir, line 123: That's fine, but we must ensure there is a separate ticket for that in the backlog

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Analysis.java
```diff
@@ -118,32 +118,37 @@
...
+    private Project project;
```
tskir, line 124: Since Project <1..1>* Study <1..M> Analysis, is there a need to explicitly fill the analysis-to-project link if we already have an analysis-to-study link? Can't the analysis-to-project be a computed property?

_\* Presumably, at least in our use case_

tskir:
So it looks like adding a support for project-based import is not that trivial. It brings a lot of new codebase along — we have to support new entity type, its repositories, converters, tests, deprecation aspects etc. This issue will probably generate a decent amount of work and follow-up tickets, which we will need to discuss and create.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
jmmut: This one was tough to figure out.

Assuming EGA will only import from DB, my understanding is that EGA will import analysis (and associated studies, but NO projects), and EVA will import projects (with studies and analysis), so I don't think getSECONDARYIDArray will make a problem here:

- **project**`.getSECONDARYIDArray` with several studies: can only have more than 1 elements for EGA, and EGA doesn't import projects.

- **project**`.getSECONDARYIDArray` with 0 studies: this too will only be called by EVA, so it will never have 0 studies.

- **study**`.getSECONDARYIDArray` with several projects: I think should never be more than 1 so I'll throw an exception. 

- **study**`.getSECONDARYIDArray` with 0 elements: In subsequent commits I allow importing a study without projects, as some tests import EGA studies that don't have projects. This code is also called from importAnalyses.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Analysis.java
jmmut: It can be a computed property. I did this because EVA will want to easily link projects and analysis without knowing the studies, but the code can allow this transparently. I will add a ticket for this.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/PublicationExtractor.java
jmmut: EVA works with projects in several places. it's going to be a constant pain having to work with studies only for calling this code.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/ObjectsImporter.java
jmmut: I created  EVA-2012 for cleaning the relationships between the entities, and we can try to simplify the logic there.

metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/entities/Analysis.java
jmmut: EVA-2012

jmmut:
> Not quite. As far as I remember, the findOrSave function from StudyRepositoryCustom is used in both DB and API import. Its purpose is to not attempt to save an object the second time if it already exists.

I really looked at this and (in the master branch) the import fails if you import a study twice, but it swallows the exception and the state is ok because it is already copied into our DB. I really think it is not used for the API import. This approach is arguably not the best but it works. I'm ok if we decide to change it.


> >    The cucumber tests about study seem about creating new studies. Do we need those for projects too?
> Apparently so — if we are introducing a new entity, we probably need to test it as extensively as all other ones.

My point was, we will create studies via ENA directly, so should we really create anything on our side ? anything we change will be overwritten if we reimport it.

>  looks like we'll have to implement DB import as well

I created EVA-2011 for that. The other comments about the relationship between project and analysis will be handled in EVA-2012.

metadata-load/src/main/java/uk/ac/ebi/ampt2d/metadata/importer/extractor/PublicationExtractor.java
tskir: Sorry, I probably didn't explain myself properly. What I proposed to do was to start import from project in EVA case; then retrieve study based on that; and then do _everything_ else starting _from the study_ which we retrieved. If we do it this way, we won't have a need for `getPublicationsFromProject` and `getPublicationsFromStudy` because the publications (and all other child objects) will be retrieved from the study. It will only affect import, and we'll be able to work with both study and project accessions in the webservice

tskir: So to clarify even further, I propose that Study becomes a central object for import for both use cases to simplify code. We will have:
* EGA case: start from Analysis → get Study → **do everything else from Study**
* EVA case: start from Project → get Study → **do everything else from Study**

tskir: I've added this to EVA-2012

